PR #39, opened 2014-09-08
Ensure file handles are closed using with statement

svisser:
This change ensures the file handles are properly closed in setup.py.


dogweather:
:+1:

PR #44, opened 2014-11-18
Use double quote to avoid broken strings

jjconti:
Use double quote to avoid broken strings in the .py file if description or author name contains a single quote.


svisser:
Shouldn't `http://cookiecutter-pypackage/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/__init__.py` be changed as well then?

PR #46, opened 2014-12-26
remove unittest.main in test file

juanpabloaj:
with `python setup.py test` , `nosetests` or `py.test` this lines are unnecessary


dogweather:
I like the spirit of this because it removes redundant, confusing ways to do something.

audreyr:
Agreed @dogweather, removing redundancy is good. Merging. Thanks @juanpabloaj.

PR #48, opened 2015-01-10
Define package version in version.py and then import that into setup.py

mw44118:
Hi -- This is a tiny tweak to make it easy to check the version of a package after it has been installed.

### DIFF
--- a/{{cookiecutter.repo_name}}/setup.py
+++ b/{{cookiecutter.repo_name}}/setup.py
@@ -19,7 +19,9 @@ test_requirements = [
     # TODO: put package test requirements here
 ]
 
-from {{ cookiecutter.repo_name }}.version import __version__
+# Read __version__ from version.py
+with open("logtopg/version.py") as f:
+    exec(f.read())
 
 setup(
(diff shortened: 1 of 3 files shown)


mw44118:
There was a bug in the previous version.

dogweather:
:+1:

eliasdorneles:
Hm, this feels too tricky to me.

I'd rather have a `.bumpversion.cfg` generated by default and recommend users to use [bumpversion](https://pypi.python.org/pypi/bumpversion) for their releasing workflow.

mw44118:
Didn't know about bumpversion.  That's interesting!

glasslion:
[Python Packaging User Guide - Single-sourcing the Project Version](https://packaging.python.org/en/latest/single_source_version/) lists a few  techniques to store the version  without duplicating the value.

Each tech has its pros & cons. We can leave a link in the setup.py and let users choose.

audreyr:
I'm playing around with bumpversion now. Thinking of closing this in favor of a generated `.bumpversion.cfg` to keep things simple.

PR #50, opened 2015-02-09
read requirements from file(s) in setup.py

spazm:
* read `requirements.txt` and `test-requirements.txt` in `setup.py`,
  skipping blank lines and comments.
* include `requirements.txt` and `test-requirements.txt` in `MANIFEST.in`
  so they are available when tox builds and tests the package
* add stub test-requirements.txt file


audreyr:
@spazm Thanks much and sorry for the delay. 

This PR is similar to #61, except that #61 does the requirements file parsing using `pip.req.parse_requirements` which saves us from having to include the actual parsing code in setup.py which is nice.

spazm:
Awesome! I'll add that to my local projects.

Issue #1, opened 2013-08-09
Don't import the package to get the version


pydanny:
Apparently, this can cause issues. Not clear how, but there you are. Here is a mitigation:

https://github.com/jezdez/envdir/blob/master/setup.py#L23-L33

PR #51, opened 2015-02-09
read version from `.__version__` in setup.py

spazm:


benjaoming:
Would be great!!! :+1: 

Someone could miss either of the two version specifications when bumping their package, which is not very nice.

audreyr:
Sorry for the delay, thanks for the pull request and feedback.

This wouldn't actually work, since the package isn't being imported in setup.py. See https://github.com/audreyr/cookiecutter-pypackage/issues/1 for a discussion of the problems with importing a package in setup.py.

PR #52, opened 2015-02-13
TST: test_{{cc.repo_name}}.py: setup, seq, exit

westurner:


audreyr:
Sorry for the delay @westurner.

Anyone else have feedback on this?

eliasdorneles:
Well, the main block using `unittest.main()` was removed in a previous commit: https://github.com/audreyr/cookiecutter-pypackage/commit/c8bd3655019a8817964ed2b3095130c8245412be
I believe the reasoning is that we shouldn't rely a main block in the test to run the test suite, it brings import issues (and then you have to fiddle with the path to make the test work as a main or install the package as editable).

Running a specific test suite can be done in different ways, with: `python setup.py test --test-suite tests.test_boilerplate` or with tools like [nose](https://nose.readthedocs.org), [py.test](http://www.pytest.org), etc.

So, +0 from me.

westurner:
> I believe the reasoning is that we shouldn't rely a main block in the test to run the test suite, it brings import issues (and then you have to fiddle with the path to make the test work as a main or install the package as editable).

I often find it justifiable to add a ``-t/--test`` CLI option to make the [program] self-testing.
(which does not require ``pip install package[dev,test]``, ``tests_require``, ``requirements-test.txt``).

* It's also necessary to pop the not-test-args from ``sys.argv`` before unittest.main.

I suppose *self-testing* is the value here, and not this little snippet.

westurner:
As a startup CRC, a ``-t/--test`` option that prempts before continuing with the next operations may or may not be necessary.

westurner:
... https://westurner.org/wiki/awesome-python-testing#test-runners

westurner:
``skipif`` would be great to have in e.g. unittest3

audreyr:
I see now, thanks for the explanation. Merged. Thank you!

PR #57, opened 2015-04-17
license: switch to ISC license

vincentbernat:
There are several BSD licenses: 4-clause (non free), 3-clause (used by
this project), 2-clause, 1-clause. Moreover, the 3-clause versions has
a non-endorsement clause where a name should be put and it is always a
bit confused on what name should be put here.

This commit switches to a simpler license, the ISC license. Only one
version of this license exists and no templating is needed for it.

That's really just a suggestion. Feel free to just discard it. I kept BSD as the license of the project. Dunno if it should be the same.


breerly:
Seems like we should allow the user to choose the license of their choice, sort of like what Github does when you create a new repo.

breerly:
Also - I think `MIT` is probably the most generic and obvious license.

pydanny:
The next version of `cookiecutter` will allow for choice of options (thanks to Click). This means we can provide a list of license options and let the generating person select their preference.

vincentbernat:
@breerly There are two versions of the MIT license. The MIT/X11 and the MIT/Expat. Usually, people go with the MIT/Expat since the MIT/X11 requires some templating. In both cases, ISC is shorter.

breerly:
We should probably just support a 1:1 of what's listed in the "Add a license" dropdown here: https://github.com/new

audreyr:
1:1 support of the "Add a license" dropdown options would be nice. Once Cookiecutter supports a choice of options, that will be possible.

breerly:
Is there a PR already out for that?

audreyr:
@breerly No, not yet.

Thanks for this @vincentbernat. I thought more about this and agree. Merging for now until we get license choice support into Cookiecutter.

PR #58, opened 2015-04-30
Using webbrowser module instead of Mac-only open to open user's browser

eliasdorneles:
Hello,
Here is a proposal for using `python -m webbrowser` to open user's browser for the commands in `Makefile`, which seems to be more portable than the Mac-only command.
Is this any good?
Thank you!

PS: btw, thanks for this awesome project, I've used to release a few packages already and it made everything so much easier. so, thank you! <3

### DIFF
--- a/{{cookiecutter.repo_name}}/Makefile
+++ b/{{cookiecutter.repo_name}}/Makefile
@@ -1,4 +1,5 @@
 .PHONY: clean-pyc clean-build docs clean
+BROWSER := python -c 'import urllib, os, webbrowser, sys; webbrowser.open("file://" + urllib.pathname2url(os.path.abspath(sys.argv[1])))'
 
 help:
 	@echo "clean - remove all build, test, coverage and Python artifacts"
@@ -47,7 +48,7 @@ coverage:
 	coverage run --source {{ cookiecutter.repo_name }} setup.py test
 	coverage report -m
 	coverage html
-	python -m webbrowser htmlcov/index.html
+	$(BROWSER) htmlcov/index.html
 
 docs:
 	rm -f docs/{{ cookiecutter.repo_name }}.rst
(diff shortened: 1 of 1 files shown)


audreyr:
@eliasdorneles Thanks for this PR and for the kind words. Really sorry about the delay. This is a nice idea, but it didn't work locally when I tried it out. I looked into it. From what I understand, the webbrowser module requires a local URL to:

* include the full absolute path
* be prefixed with `file://`

I'm not actually sure if it's possible to do both of the above when executing `webbrowser` as a script. Do you or anyone know?

eliasdorneles:
@audreyr hm, right, behavior for file paths seems to be dependant of OS.

`python -m webbrowser htmlfile_path` works for me on Chrome/Linux Mint.

I've just sent an update trying something new: a python one-liner that attempts to create a proper URL given a relative file path and open it.
Does this work for you?

Thank you!

audreyr:
Thanks for the update!

Got this error on Python 3.4:

```
python -c 'import urllib, os, webbrowser, sys; webbrowser.open("file://" + urllib.pathname2url(os.path.abspath(sys.argv[1])))' docs/_build/html/index.html
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: 'module' object has no attribute 'pathname2url'
make: *** [docs] Error 1
```

Looks like in Python 3, the path is now `urllib.request.pathname2url`

eliasdorneles:
hm, I'm struggling to find a way to use this Python code in a Makefile:

```
import os, webbrowser, sys
try:
       from urllib import pathname2url
except:
       from urllib.request import pathname2url
webbrowser.open("file://" + pathname2url(os.path.abspath(sys.argv[1])))
```

will try again a bit more tomorrow!

eliasdorneles:
Some generous soul helped me to solve the Make issue: http://unix.stackexchange.com/a/222917/12954

@audreyr can you please try again now? =)

eliasdorneles:
hey @audreyr, do you have any feedback on this approach?
thank you!

audreyr:
Thanks for pinging me. This is awesome, nice work!

PR #61, opened 2015-05-25
Read requirements from requirements.txt and add to install_requires

kracekumar:
It is possible to use this layout for private repo/or add git link
to requirements or some one can run `python setup.py develop`.
In those case all requirements should be installed.

### DIFF
--- a/{{cookiecutter.repo_name}}/setup.py
+++ b/{{cookiecutter.repo_name}}/setup.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+from pip.req import parse_requirements
+from pip.download import PipSession
 
 try:
     from setuptools import setup
@@ -14,7 +16,9 @@ with open('README.rst') as readme_file:
 with open('HISTORY.rst') as history_file:
     history = history_file.read().replace('.. :changelog:', '')
 
-requirements = [
+pip_reqs = parse_requirements("requirements.txt", session=PipSession())
+
+requirements = pip_reqs + [
     # TODO: put package requirements here
 ]
 


audreyr:
@kracekumar This is nice, thanks. What minimum version of pip does this require?

benjaoming:
:+1:

benjaoming:
Hmm, this makes pip a pre-requirement for running setup.py, not sure if that's wise?

Why not just dump the raw contents of requirements.txt without parsing them?

benjaoming:
I kind of like #50 better to be honest... it saves importing pip and doesn't make the parsing any harder than necessary.

pydanny:
I may agree with @benjaoming's concerns. As far as I know, no Python 2.7 distribution comes with pip built in. Hence, we possibly need to test this across Homebrew and various Linux package implementations.

kracekumar:
@audreyr AFAIK the functionality is available since 1.5. Changelog or docstring doesn't clearly say when it was introduced, but simple test in interpreter confirms it.

The #50 fails for some thing like

```
# requirements.txt
-r base.txt
uwsgi==2.1.0
```

kracekumar:
The pull request can be modified to work if pip is available else fall back to original implementation.

benjaoming:
@kracekumar true, but that pattern is also sort of odd and requires pip anyways... so the idea would be to avoid encouraging people to write pip-requiring setup.py's, yet to put dependencies in requirements.txt because it's easy to read.

I don't like the idea of loading in a new pip session for parsing two lines of simple requirements.. in 99% of cases it would be overkill :)

benjaoming:
But I agree with the intention! It's really bad to have requirements specified in two different places!

eliasdorneles:
Hm, I feel I should give my 2 cents here.

Making install_requires == requirements.txt by default may cause users of your package to downgrade/upgrade libraries unnecessarily according to your requirements.txt

The way I see it, the dependencies in `requirements.txt` and the dependencies in `install_requires` in `setup.py` represent different things: the former is meant for a repeatable development environment, and the latter represents the very minimal dependencies needed at runtime.

So, in `requirements.txt` we put the dependencies that we want to use in development environment (including optional dependencies) and we make them very specific (e.g., using `==VERSION`) to ensure that every developer will use the same environment.

And in `install_requires` we declare only the minimal dependencies and we make them very loose (e.g., using `>=VERSION` the most ancient version known to work) in order to avoid user having to downgrade or upgrade a package unnecessarily.

So yes, it's true that we end up declaring most dependencies twice, but there are good reasons for that.

benjaoming:
@eliasdorneles 
Thanks for the explanation, but I have to disagree.

A common pattern for me would be:

**requirements.txt**:

```
some-req==1.2.3
```

**requirements_dev.txt**:

```
-r requirements.txt
debug-foo==1.2.3
```

Usually, we put the *minimal runtime dependencies* in requirements.txt and have an explicitly named file if there's additional dependencies for development.

In this pattern, we would also read requirements.txt in setup.py.

eliasdorneles:
@benjaoming hm, not sure if we're talking about the same thing here.

I wasn't talking about simply optional packages used in development (debugging and test running stuff). The differentiation I was referring to was about _version_ of the actual runtime dependencies of your package.

Say, if you put in your setup.py `some-req==1.2.3`, and an user has `some-req==1.2.4` (a more up-to-date version than what's declared in your `setup.py`), when this user installs your package, it will force that specific version, causing your user to uninstall the more up-to-date version and install the older one that is unnecessarily pinned in your `setup.py`.

That's what I was referring to as unnecessary downgrading -- we want to avoid that.

But we also want our development environments repeatable, so we do want to pin the specific versions in `requirements.txt`.

So, the thing to do is to put, in `requirements.txt`:

    some-req==1.2.3

And in `setup.py`:

    install_requires=[
        'some-req >= 1.2.0', # or whatever is the needed minimal version
    ]

I hope my point is clearer now. :)

benjaoming:
@eliasdorneles 
> That's what I was referring to as unnecessary downgrading -- we want to avoid that.

Agreed, but do you think a project should ship with a requirements.txt full of over-specified version numbers? Why not ship with a requirements.txt that's exactly what's required and then pin things in a requirements_test.txt ?

It's both non-DRY and non-explicit to have a requirements.txt that *isn't* used in setup.py.

eliasdorneles:
> Agreed, but do you think a project should ship with a requirements.txt full of over-specified version numbers?

Yes, because:

1) it's about making development environments repeatable (everyone testing on the same versions).
2) it's what many people do by default and expect to continue doing (many just do: `pip freeze -l > requirements.txt`)

> Why not ship with a requirements.txt that's exactly what's required and then pin things in a requirements_test.txt

Hm, wouldn't this have the same issues with DRY you were concerned?

I'd rather have the defaults being cleaner (no tricks needed to read requirements.txt from setup.py, hence minimizing chance of installation issues, no extra file to pin versions).

benjaoming:
> it's what many people do by default and expect to continue doing (many just do: pip freeze -l > requirements.txt)

No that's not what people do necessarily. *Some* do that. But I don't buy that requirements.txt is a conventional file used for test requirements: That's why this PR exists!

So the discussion is about if you find it more natural that requirements.txt is automatically used in setup.py or if it should be allowed to differ. IMO it's better to have a requiements_test.txt or similar explicit convention if you want pinned versions for testing. DRY isn't broken by this, because your requirements.txt will be different when it's not pinning versions. If pip allowed to overwrite requirements imported from another file with `-r`, it would make things a little easier.

Also, on a minor side-node. It's dangerous to pin things with no upper boundaries, because the dependencies that you ship with will be broken in later releases. Small example: django-cms 2.3 [shipped with a dependency of django-classy-tags without an upper bound](https://github.com/divio/django-cms/blob/support/2.3.x/setup.py#L51), so if you deploy django-cms 2.3 now, it won't work because it installs an incompatible django-classy-tags 0.4.

kracekumar:
@audreyr Please say if this relevant or not. I will close this if it is not relevant. I feel we're diverging from original point.

eliasdorneles:
Btw, this is an interesting related article: https://caremad.io/2013/07/setup-vs-requirement/

@benjaoming maybe using `-e .` in your requirements.txt as suggested in that article would solve your issue. :)

benjaoming:
@eliasdorneles yes definitely using `-e .` is DRY and looks good! Nice! :) But it contradicts pinning versions in requirements.txt as you mentioned earlier... so you'd still need something like a requirements_test.txt  with the pinned versions for testing or deployment.

eliasdorneles:
@benjaoming yeah, personally I wouldn't use it, I like the separation I mentioned. I figured you'd like it, though! =)

audreyr:
Thanks everyone for the discussion, this has been helpful. I'm still trying to digest everything and come to the right conclusion...

audreyr:
After much thought, I've decided to close this.

I'll be renaming the current requirements.txt to requirements_dev.txt, since bumpversion and wheel are just dev requirements. When setup.py supports `dev_requires`, I may replace requirements_dev.txt with that.

Note: this means there won't be a requirements.txt file provided for version pinning; however, one can create it after generation if they really want to track pinned versions.

eliasdorneles:
Neat! :+1:

benjaoming:
@audreyr looking forwards to see your next work of unifying best practices!! :+1:

csala:
@audreyr I know that this is already closed, but I would like to respond to your last comment:

> When setup.py supports dev_requires, I may replace requirements_dev.txt with that.

`setup.py` does not support `dev_requires` yet, but it's because there is an alternative which seems to be an accepted practice, which is using the `extras_require`, like so:

```
requirements = [
    # actual package requirements
]
dev_requirements = [
    # bumpversion or twine might go here
]

setup({
    install_requires=requirements,
    extras_require={
        'dev': dev_requirements
    }
})
```

Then, you can install the package alongside all the requirements needed for development using `pip install -e .[dev]`.

I recently asked about this in the `distutils` mailing list and there seemed to be a lot of consensus around this practice: https://mail.python.org/mm3/archives/list/distutils-sig@python.org/thread/OQW4VU36WWGESOAYLTV6WWDKX2RKHRD4/

Perhaps it would be a good option to add to the cookiecutter?

PR #64, opened 2015-09-21
Travis setup for automated pypi deployment

eliasdorneles:
Hey @audreyr,

So, here's a sketch of how the Travis setup would look like for an automated PyPI deployment setup.

This change would make Travis deploy the project using the username and password configured, whenever there a new tag is pushed to the Git repo, except I got stuck trying figure out how to implement updating `.travis.yml` with the encrypted password.

Problem: we can only encrypt the password after the Github repo exists and is configured in Travis.

The password encryption work in two steps:

1) Get the public RSA key from Travis project for the repo configured, through an URL like this: `https://api.travis-ci.org/repos/audreyr/cookiecutter/key`: note that this will only work if the Github repo already exists and is added to Travis!
2) Use that key to encrypt the pypi user password, and put it in `.travis.yml` in the correct place (where now is the string `PLEASE_REPLACE_ME`): this bit is straightforward, can be done in [a few lines of pure Python code](https://github.com/sivel/travis-encrypt/blob/master/python/travis_encrypt.py#L61).

So, it's kind of a chicken-and-egg problem, and I'm kinda lost.

As far as I can see, there are a few options:

1. Put the plain-text user PyPI password in the `.travis.yml` file -- this would work, but it would obviously be a huge security hole and I'm only writing it here for the sake of completeness.

2. Provide a script that the user can run after creating the Github repo and adding it to Travis for updating the .travis.yml (we can print a reminder for this on a post-generate hook)

3. Leave the whole deploy section commented-out in the `.travis.yml` file, with instructions for how to proceed to enable it

Maybe there are other options that I'm missing, I dunno.

Thoughts?


audreyr:
Very cool! I think option 2 or 3 would be fine. I like 2 better as long as its usage is also documented in the readme or somewhere.

This is not necessary for this PR, but just a thought for the future: I wonder if we could automate creating the Github repo and adding it to Travis. Awhile back, I played with the repo creation part by using the Github API in a post-generate hook (https://github.com/audreyr/cookiecutter-github/blob/master/hooks/post_gen_project.py). Perhaps we could do the same with Travis. Then we could make it optional via an option at the prompt.

eliasdorneles:
Wow, that would be the ultimate automation! :)
I was taking a look at Travis API docs and I think it's possible, it seems quite complete.

Okay, so I'll implement option 2 for this PR and we can build the ultimate automation separately since it will need some further digging.

Btw, I have the feeling that propagating the cookiecutter template variables to the hook scripts as environment variables would make it easier, I wonder how you feel about that. =)

eliasdorneles:
Hey @audreyr,

So, here is an updated proposal for the option 2 discussed above, adding the script to update Travis config file and updated the instructions for next steps in README.

There is relatively minor side effect when updating the `.travis.yml` file via the script: the comments are lost and previous ordering isn't preserved (mainly because yaml are like Python dicts).
This is the resulf of using the [PyYAML](http://pyyaml.org/wiki/PyYAML) library.

I think options here are:

1. Implement a state-based line-oriented algorithm (like a sed script) to update the file preserving the previous content.

2. Try other YAML libraries -- there is [at least one](https://pypi.python.org/pypi/ruamel.yaml) that attempts to handle comments and ordering in YAML, it doesn't seem as battle-tested as PyYAML and bring other dependencies too, but maybe that's okay.

3. Just leave the current behavior. If we think comments or ordering in `.travis.yml` are important, we can warn the user beforehand or add an option to the script to print to the screen instead of writing to the file.

Implementing option 1 would be fun (I actually started it), but the resulting code was ugly and looks like a source of bugs, so I just skipped it.

PS: I left PyYAML out of package's requirements.txt because it isn't a runtime dependency (it's only needed for the script), let me know if you prefer it another way.

Let me know what you think! :)

pydanny:
@eliasdorneles, here is my quick 2 cents:

1. Option 3 (leave the current behavior) sounds good to me. I would just inject a warning comment into the YAML after creation saying, "This YAML file is autogenerated and will overwrite each time it's deployed."

2. Once we have option 3 in place, we can explore the other options. Lessons learned from implementing option 3 will really help in researching them.

2. `Crypto` is not part of the standard library. If it is added, then an option should be placed into `cookiecutter.json` specifying whether or not they want the travis deployment. Here's an example of controlling the `post_gen_hook`: https://github.com/pydanny/cookiecutter-django/blob/master/hooks/post_gen_project.py#L111-L113

3. If a third-party encryption library is going to be used, I strongly recommend using [cryptography](https://github.com/pyca/cryptography) instead.

eliasdorneles:
Oh my, you're right, I don't know why I concluded `Crypto` was stdlib! Sorry for the confusion, I think I got overexcited. :)

Thanks for reviewing, all great points. Nice trick on `post_gen_hook` -- I didn't know it was possible to use template variables in the hook scripts, that solves my concern from previous comment. :+1:

Thanks for pointing to cryptography library, I'll dig into it and send an update later. :)

eliasdorneles:
Hey there @pydanny @audreyr,

I applied the changes according to the review: the script now uses cryptography module instead of PyCrypto and puts a small comment warning of possibility of overwriting (which in most cases should happen only once, at setup time), added a flag to cookiecutter.json offering the option to not use the pypi deployment with Travis and also put dev requirements in the `requirements_dev.txt` file.

I also rebased the changes on top of current master.
Build isn't passing because of issue fixed in PR #66.

Does this look good now? :)

hooks/post_gen_project.py
```diff
@@ -0,0 +1,14 @@
...
+
+PROJECT_DIRECTORY = os.path.realpath(os.path.curdir)
+
+
+def remove_file(filepath):
+    os.remove(os.path.join(PROJECT_DIRECTORY, filepath))
+
+
+if __name__ == '__main__':
+    if '{{ cookiecutter.use_pypi_deployment_with_travis }}'.lower() == 'n':
```
pydanny: If `'{{ cookiecutter.use_pypi_deployment_with_travis }}'.lower() == 'y':`, then this hook will break the building of the package because `if __name__ == '__main__':`'s code block will lack even a pass.

eliasdorneles: uh, how do you mean?
Sorry, I'm not following what's the issue here, could you please give more details?

If user replies `'n'`, this removes the file, if user replies `'y'` (the default), it doesn't do nothing.
The `if __name__ == '__main__'` block is never empty.

Thanks for reviewing! :+1:

pydanny:
@eliasdorneles In retrospect, I read that bit of code incorrectly. Such is the abstraction of post-hooks in cookiecutter.

{{cookiecutter.repo_name}}/requirements_dev.txt
```diff
@@ -1,2 +1,4 @@
 bumpversion==0.5.3
 wheel==0.23.0
+cryptography==1.0.1
+PyYAML==3.11
```
audreyr: Let's wrap these 2 lines in `{% if cookiecutter.use_pypi_deployment_with_travis == 'y' -%}`

{{cookiecutter.repo_name}}/.travis.yml
```diff
@@ -1,16 +1,32 @@
...
 
 # command to install dependencies, e.g. pip install -r requirements.txt --use-mirrors
-install: pip install -r requirements.txt
+install: pip install -U tox
 
 # command to run tests, e.g. python setup.py test
-script: python setup.py test
+script: tox
+
+{% if cookiecutter.use_pypi_deployment_with_travis != 'n' -%}
```
audreyr: Better to say `== 'y'` to avoid double negative

{{cookiecutter.repo_name}}/travis_pypi_setup.py
```diff
@@ -0,0 +1,89 @@
...
+except:
+    from urllib.request import urlopen
+
+
+GITHUB_REPO = '{{ cookiecutter.github_username }}/{{ cookiecutter.repo_name }}'
+TRAVIS_CONFIG_FILE = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)), '.travis.yml')
+
+
+def encrypt(pubkey, password):
```
audreyr: Could you please add docstrings to this and the other functions, to help me understand better? :)

{{cookiecutter.repo_name}}/requirements_dev.txt
eliasdorneles: good catch, thanks!

{{cookiecutter.repo_name}}/travis_pypi_setup.py
eliasdorneles: oh, how inconsiderate of me! :)
sure thing, give me a moment!

{{cookiecutter.repo_name}}/.travis.yml
eliasdorneles: Alright, so I'll change the test in the hook to check with `!= 'y'` too, to ensure consistency.
Namely, we'll only apply this if user doesn't change default answer `y`, and we don't apply in any other case.
Let me know if this doesn't sound good.

eliasdorneles:
I've just noticed that the PyYAML is converting the `on:` from the original `travis.yml` to `true` -- gimme a few days, I'll do a bit more testing to see if this can be a problem with Travis.
PS: I'll also rebase and solve the conflicts.

eliasdorneles:
Hey @audreyr!

Regarding my comment above, I've verified that the rewrite from `on:` to `true:` is not a problem.
I tested this on a [series of commits on a test repo](https://travis-ci.org/eliasdorneles/citesting/builds) -- it seems that [`on` and `true` kinda equivalent in YAML](http://yaml.org/type/bool.html), I didn't know about that. =)

I've rebased the PR and also did some refactoring and added docstrings to the script, aiming to make it more readable. Let me know if it looks better now and if there is anything I can make more clear.

Thanks for reviewing!

audreyr:
Looks great @eliasdorneles, merging.

PR #65, opened 2015-09-25
Add make servedocs to watch for changes and recompile docs

eliasdorneles:
Hey @audreyr,

This adds a `make servedocs` target that does what `make docs` do and then watch for changes in the docs recompiling whenever you save.
It uses the `watchmedo` utility from the [watchdog](https://pypi.python.org/pypi/watchdog) package available in PyPI.

I've been using it in a few projects for a while, it has proved handy. :)

Does this look good?

Thanks,
Elias

### DIFF
--- a/{{cookiecutter.repo_name}}/Makefile
+++ b/{{cookiecutter.repo_name}}/Makefile
@@ -68,6 +68,9 @@ docs:
 	$(MAKE) -C docs html
 	$(BROWSER) docs/_build/html/index.html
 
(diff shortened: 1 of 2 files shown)


audreyr:
This is a good idea.
Can you add watchdog to `requirements_dev.txt`?

eliasdorneles:
@audreyr sure thing, it's done! :+1:

audreyr:
Thanks @eliasdorneles!

PR #66, opened 2015-09-26
Fix build and simplify Travis config

eliasdorneles:
Hey @audreyr,

This fixes the build and cleans up `.travis.yml`: since boilerplate requirements file is only dev requirements, we can assume boilerplate has no runtime deps and so no need to install them anymore.

Looks good?


audreyr:
Appreciate this, thanks @eliasdorneles!

PR #68, opened 2015-10-01
Add cookiecutter template tag

hkraal:
I assume this is a harcoded path which went unseen.


audreyr:
Hi @hkraal, this is actually deliberate. The way it works is:
1. The package is generated by `cookiecutter . --no-input`, resulting in the `boilerplate/` directory.
2. Then the test suite is run from inside of `boilerplate/`.

PR #69, opened 2015-10-15
Fixed .travis.yml with env instead of python

kevinastone:


audreyr:
Thanks much @kevinastone.

PR #71, opened 2015-10-22
Comment deps section in tox.ini and add note about pinning

### LINKED ISSUE #70: tox.ini references non-existent requirements.txt
https://github.com/audreyr/cookiecutter-pypackage/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/tox.ini#L9

eliasdorneles:
Hey @audreyr 

This fixes #70.

I thought it was a good idea to leave the section commented with a note there, for folks who will pin the versions later.

Does this look good?


audreyr:
Perfect, thanks @eliasdorneles!

PR #72, opened 2015-10-23
Change double quotes to single quotes in setup.py

nickfrostatx:


eliasdorneles:
Uh, what's the motivation for this change?

pydanny:
Maybe because they are used in https://github.com/audreyr/cookiecutter-pypackage/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/tests/test_%7B%7Bcookiecutter.project_slug%7D%7D.py?

eliasdorneles:
Seems this is really just a minor style thing -- no behavior changes, no actual need.
+0 from me.

Issue #73, opened 2015-11-22
Error with travis_pypi_setup.py


thornomad:
Hi there - am trying this cookiecutter package out for the first time.  Thanks in advance!

I created initialized my repository, pushed it to github, and created/linked it with travis-ci.  I even pushed another build to get travis-ci to run the tests once (they failed but I wanted to make sure it was linked).  

When I runn `travis_pypi_setup.py` I am getting this error:

    ~$ ./travis_pypi_setup.py
    Traceback (most recent call last):
      File "./travis_pypi_setup.py", line 122, in <module>
        main(args)
      File "./travis_pypi_setup.py", line 107, in main
        public_key = fetch_public_key(args.repo)
      File "./travis_pypi_setup.py", line 61, in fetch_public_key
        data = json.loads(urlopen(keyurl).read())
      File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/json/__init__.py", line 312, in loads
        s.__class__.__name__))
    TypeError: the JSON object must be str, not 'bytes'

Is there something else I need to do?  Do I need my own PGP key ? 

Appreciate it -- am sure I am missing a step.

eliasdorneles:
Hey @thornomad -- thanks for reporting!

It seems I forgot to test properly the script in Python 3, sorry about that!
I'll test it properly and send a PR soon, but I think you can solve the problem by updating the `travis_pypy_setup.py` script, in the function `fetch_public_key`, replace the line:

        data = json.loads(urlopen(keyurl).read())

for:

        data = json.loads(urlopen(keyurl).read().decode('utf-8'))

The reason is that in Python 3, `urlopen(URL).read()` returns bytes(), and we need to decode them into `str` objects before passing it to `json.loads` function.

Please let me know if this helps you or not.
Thanks!

thornomad:
Yes, that fixed it that part ... Thanks.

However, after I enter my PyPI password I receive this error:

    Traceback (most recent call last):
      File "./travis_pypi_setup.py", line 122, in <module>
        main(args)
      File "./travis_pypi_setup.py", line 109, in main
        update_travis_deploy_password(encrypt(public_key, password))
      File "./travis_pypi_setup.py", line 51, in encrypt
        encrypted_password = key.encrypt(password, PKCS1v15())
[...]

PR #74, opened 2015-11-22
Fix travis_pypy_setup script for PY3, add tests for it

eliasdorneles:
So, this fixes the problems reported in issue #73 when running it in Python 3 and adds tests for the `travis_pypy_setup.py` script via tox too.

The test script is probably not perfect, but will help avoiding regressions at least for the happy path. :)


eliasdorneles:
I can't seem to be able to get a pypy 2.6 working in Travis... :/

I'll just disable testing on it for now.

audreyr:
Looks great @eliasdorneles.

PR #75, opened 2015-11-25
Switchable py.test support

kiivihal:
I have added an option to the `cookiecutter.json` to switch between py.test and the regular test setup. I have been using this for my own projects and it seems to be a often heard feature request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,15 @@ nosetests.xml
 .project
 .pydevproject
 
+# IntelliJ Idea family of suites
+.idea
+*.iml
+## File-based project format:
+*.ipr
+*.iws
+## mpeltonen/sbt-idea plugin
+.idea_modules/
+
 # Complexity
 output/*.html
 output/*/index.html
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,8 @@ install:
 
 script:
   - cookiecutter . --no-input
+{% if cookiecutter.use_pytest == 'y' -%}
+  - cd ./python_boilerplate && py.test
+{% else %}
   - cd ./python_boilerplate && python setup.py test
+{% endif %}
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,7 @@ cookiecutter-pypackage
 Cookiecutter template for a Python package. See https://github.com/audreyr/cookiecutter.
 
 * Free software: BSD license
-* Vanilla testing setup with `unittest` and `python setup.py test`
+* Vanilla testing setup with `unittest` and `python setup.py test` or `py.test`
 * Travis-CI_: Ready for Travis Continuous Integration testing
 * Tox_ testing: Setup to easily test for Python 2.6, 2.7, 3.3, 3.4
 * Sphinx_ docs: Documentation ready for generation with, for example, ReadTheDocs_
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -9,5 +9,6 @@
 	"pypi_username": "{{ cookiecutter.github_username }}",
 	"year": "2015",
 	"version": "0.1.0",
+	"use_pytest": "n",
 	"use_pypi_deployment_with_travis": "y"
 }
--- a/{{cookiecutter.project_slug}}/CONTRIBUTING.rst
+++ b/{{cookiecutter.project_slug}}/CONTRIBUTING.rst
@@ -77,7 +77,7 @@ Ready to contribute? Here's how to set up `{{ cookiecutter.project_slug }}` for 
 5. When you're done making changes, check that your changes pass flake8 and the tests, including testing other Python versions with tox::
 
     $ flake8 {{ cookiecutter.project_slug }} tests
-    $ python setup.py test
+    $ python setup.py test or py.test
     $ tox
 
    To get flake8 and tox, just pip install them into your virtualenv.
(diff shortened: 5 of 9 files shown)


eliasdorneles:
I like py.test and I like the overall idea of this PR. :)

However, I think we should use tox -- so perhaps the PR #74 should be merged first.

Also, there are a lot of conditionals here, which makes an extra hurdle for maintenance.

Personally, I'd prefer to switch to py.test already for running the tests, but keeping the test suite using only unittest -- this way people that prefer to use nosetests or some other test running tool can switch easily.

In any case, I think it would be nice to avoid the dependency at code level -- keeping it only a runtime dependency.

@audreyr wondering how do you feel about this. :)

kiivihal:
@eliasdorneles I think that is a very good solution to use py.test but keep the test suit using unittest. 

I can update the pull-request to remove the conditionals, if we agree this is the bes away forward.

bionikspoon:
One way to limit the maintenance overhead that you mentioned:  create a `test` task in `setup.py`.  Then anywhere that needs to run a test (from tox, travis, to references in the docs) should call "`python setup.py test`".  This way the conditional logic needs to be in one single place: `setup.py`; or 2 place--the cookie cutter scaffolds out a sample test, so in the test folder as well.  

I use a pytest task in my own fork: [setup.py:L13](https://github.com/bionikspoon/cookiecutter-pypackage/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/setup.py#L13).  It's wired in [line 63](https://github.com/bionikspoon/cookiecutter-pypackage/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/setup.py#L63) useing the `cmdclass` argument of `setup.py`.  

That test cmdclass could equally be a testrunner for unittest or even nose.

audreyr:
Sorry for the delay! I actually like the idea of switching entirely to py.test, both to run the tests and to replace unittest. 

I agree that the conditionals add to the maintenance burden; I think if someone really wants to use unittest, they can do so in their fork.

eliasdorneles:
Thanks for the blessing, @audreyr!
I'd still keep the unittest based suite, though -- seems a cheap way of supporting more scenarios, I'd say we have more to gain than to lose. =)

Hey @kiivihal, could you sync and update the PR? :)
Let me know in case you won't have time for it, I can work on a follow-up.
Thank you!

audreyr:
@eliasdorneles Just to clarify, do you mean:
1. Remove the conditionals
2. Use py.test to run the tests
3. Use unittest in {{cookiecutter.project_slug}}/tests/test_{{cookiecutter.project_slug}}.py

I'm curious, what is the scenario where someone would want to use py.test for the test runner but write their first package tests using unittest anyway? 

I'd still prefer to just stick with one or the other, either py.test or unittest. But I could be convinced if there's a common scenario where this is needed.

eliasdorneles:
@audreyr right, that's what I meant -- I've used pytest a few times just for the test running features (e.g. running doctests, colorized output, pytest.ini, etc), with no test code depending on pytest library itself (no `import pytest`).

However, this may be overly conservative from my part, I don't have any strong feelings strongly about, moving fully to py.test is cool too! :)

kiivihal:
I can update the pull-request to switch fully to py.test. There are two
main ways to write unit test in py.test. The classic XUnit style (
https://pytest.org/latest/xunit_setup.html) and the preferred function
based styles with fixtures. I propose that we include a best-practice
example of both, so that the user can choose which style they prefer. This
would also include a basic setup of conftest.py that is used for fixtures
that are available in multiple modules. For someone new to py.test this
would make it much easier to have a running start writing unit-tests
without having to go through the py.test documentation how to set this up.

So concretely I will do the following:


   - remove the conditional boiler plate
   - run all test (setup.py and tox) with py.test
   - remove the unittest example
   - replace it with best practice setup of py.test test files

Is this ok?


On Sun, Dec 13, 2015 at 11:46 PM, Elias Dorneles <notifications@github.com>
wrote:

> @audreyr <https://github.com/audreyr> right, that's what I meant -- I've
> used pytest a few times just for the test running features (e.g. running
> doctests, colorized output, pytest.ini, etc), with no test code depending
> on pytest library itself (no import pytest).
>
> However, this may be overly conservative from my part, I don't have any
> strong feelings strongly about, moving fully to py.test is cool too! :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/audreyr/cookiecutter-pypackage/pull/75#issuecomment-164307405>
> .
>



-- 
Sjoerd Siebinga
Delving B.V., Rotterdam
http://www.delving.eu
skype: sjoerdsiebinga
mobile: +31681926731

eliasdorneles:
@kiivihal sounds good to me!

audreyr:
:+1:

touilleMan:
Hi,

Any improvement on this ?

benjaoming:
Seems to be broken... has repo_name been removed?

```
Unable to create file 'tests/test_{{cookiecutter.project_slug}}.py'
Error message: 'dict object' has no attribute 'repo_name'
Context: {
    "cookiecutter": {
        "email": "asdasdasd@asdasdsad.asda", 
        "full_name": "asdasd", 
        "github_username": "benjaoming", 
        "project_name": "djangongo", 
        "project_short_description": "Bootstrap Django for small, flat organizations with mailman3", 
        "project_slug": "djangongo", 
        "pypi_username": "benjaoming", 
        "release_date": "2016-05-15", 
        "use_pypi_deployment_with_travis": "y", 
        "use_pytest": "y", 
        "version": "0.1.0", 
        "year": "2016"
    }
}
```

eliasdorneles:
Thanks for the fix, @benjaoming ! :+1:

PR #76, opened 2015-11-26
Add a forked version.

ardydedase:
Changes to the forked version:

* Required packages are not hardcoded in the ``setup.py`` file. All the required packages are inside the ``requirements`` folder.
* Package requirements are broken down into separate files::

.. code:: bash

    ├── requirements
    │   ├── dev.txt
    │   ├── prod.txt
    │   ├── test.txt

* Creates a default class based on the package name.


audreyr:
Thanks @ardydedase.

PR #78, opened 2015-12-10
Add python 3.5 to contributing guide

kragniz:
The tox environment specifies that we should test 3.5, so this guide should also say the same.


eliasdorneles:
Makes sense, thanks @kragniz !

kragniz:
:sparkles:

Issue #80, opened 2015-12-12
remove ReST `:changelog:` directive.


bionikspoon:
What does the `.. :changelog:` directive do?  I could not find it anywhere in the docs for Sphinx or docutils.  It's not in docutils node list, or any reasonable places in Sphinx' source (`sphinx/directives/`, `sphinx/domains/`).  I browsed readthedocs, didn't see it there. 

It seems the only thing it does for certain is break pypi's ability to render rst.

Could I suggest removing it?

audreyr:
This was a workaround that @pydanny added to get the files to render as RST properly on PyPI. Do the relevant sections display on PyPI properly without it?

bionikspoon:
@audreyr That directive was actually being removed from the PyPI description page in the setup.py script.  PyPI uses a very strict rst interpreter that breaks on anything non standard.

I'm glad to see this is safe to remove.  Thanks @audreyr and @pydanny 

ref: #81

PR #81, opened 2015-12-12
Remove changelog directive. #80

bionikspoon:


audreyr:
This is needed in order to display properly on PyPI. (Correct me if you have a better PyPI workaround, though.)

audreyr:
Actually, @pydanny says it might not be needed anymore? Can you please verify?

pydanny:
It was hackery on my part from a while back, and hasn't been needed in a while.

audreyr:
Thanks @bionikspoon and @pydanny :thumbsup:

PR #89, opened 2016-02-24
Console script

ramakrishna61926192:


eliasdorneles:
hello @ramakrishna61926192 -- it's not very polite to create a PR for code that you haven't contributed yourself. :)

Having said that, I hadn't noticed the console-script branch before. :)
I kinda like it, but for merging it I'd expect it to be optional and the script name to be configurable, users probably wouldn't want to use `main` as a cmdline tool name.

@audreyr is that a desired feature, do you think? If you still would like this feature, I can try to follow up on it.

pydanny:
I think these changes remained in a branch because @audreyr hadn't completed it yet. Or was waiting for features to be implemented in Cookiecutter itself.

IMO, the console script functionality has components that should be managed in a post hook.

eliasdorneles:
You mean, things like removing the script if user is not interested in having a cmdline script entrypoint?

audreyr:
I think supporting console scripts in the main branch would be nice as a configurable option. If either of you feel like implementing this (or anyone else who happens to read this), go for it.

About that branch: originally I was using it with Cookiecutter's branch checkout feature whenever I wanted to create a console script package. But then I forgot about it and never merged master into it. It's probably better off deleted and supported as an option in the main branch.

PR #91, opened 2016-03-06
Fixed travis builds with py3.5

likeon:


{{cookiecutter.project_slug}}/.travis.yml
```diff
@@ -15,7 +16,7 @@ env:
 install: pip install -U tox
 
 # command to run tests, e.g. python setup.py test
-script: tox
+script: tox -e $TOXENV
```
eliasdorneles: is this line really necessary?

likeon: Otherwise, tox runs tests for every environment multiple times.  I checked it in my project; you can find it here https://travis-ci.org/likeon/python-headhunterru/builds/114120373.

eliasdorneles: I believe the issue is that your project is using `TOX_ENV` instead of [`TOXENV`](http://tox.readthedocs.org/en/latest/example/general.html#selecting-one-or-more-environments-to-run-tests-against).

likeon: Looks like you right, -e option isn't necessary here. I found the solution in py3.5 builds in your cookiecutter's .travis.yml that uses TOX_ENV. Pushed the fix. Thank you.

eliasdorneles:
Thanks @likeon !

PR #92, opened 2016-03-07
Include images in documentation

miohtama:
The docs may contain related images which were ignored by MANIFEST.in


audreyr:
This is a good catch, thanks @miohtama

PR #97, opened 2016-05-01
Proper highlight shell commands (fixes #94)

### LINKED ISSUE #94: Using code-block in Sphinx by default
Does the default installation of Sphinx include Pygments for syntax highlighting (or should this be supported).

This way one could do:

        .. code-block:: console

And 

      .. code-block:: python

In the default usage.rst and installation.rst to get proper syntax highlighting.

eliasdorneles:
Add proper highlight support for shell commands.

### DIFF
--- a/{{cookiecutter.project_slug}}/docs/installation.rst
+++ b/{{cookiecutter.project_slug}}/docs/installation.rst
@@ -4,11 +4,15 @@
 Installation
 ============
 
-At the command line::
+At the command line:
+
+.. code-block:: console
 
     $ easy_install {{ cookiecutter.project_slug }}
 
-Or, if you have virtualenvwrapper installed::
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/docs/installation.rst
```diff
@@ -4,11 +4,15 @@
 Installation
 ============
 
-At the command line::
+At the command line:
+
+.. code-block:: console
 
     $ easy_install {{ cookiecutter.project_slug }}
```
eliasdorneles: Btw, @audreyr, how do you feel about changing the default advice to recommend just `pip install {{ cookiecutter.project_slug }}` by default?

audreyr: I think that's a good idea.

PR #98, opened 2016-05-07
Add more complete installation instructions

eliasdorneles:
Proposing more complete installation docs, suggesting pip install for
the stable releases and `python setup.py install` for installation from
sources.

For people that don't have pip, we point them to the Python installation in the hitchhiker's guide, that has nice instructions for having the basics (setuptools, pip and virtualenv): http://docs.python-guide.org/en/latest/starting/installation/

### DIFF
--- a/{{cookiecutter.project_slug}}/docs/installation.rst
+++ b/{{cookiecutter.project_slug}}/docs/installation.rst
@@ -4,15 +4,46 @@
 Installation
 ============
 
-At the command line:
+
+Stable release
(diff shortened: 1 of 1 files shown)


eliasdorneles:
Quick preview:
selection_2016-05-07_12 22 23

audreyr:
This is great, @eliasdorneles.

eliasdorneles:
glad you liked it. :)

PR #100, opened 2016-05-15
Do not use removed repo_name in PyTest template

benjaoming:
`repo_name` does not exist.

Don't know if changing from `capitalize` to `title` has a point, but it seems to work fine.


benjaoming:
Using py.test doesn't work, was merged yesterday in #75

eliasdorneles:
Looks good, thanks @benjaoming !

PR #101, opened 2016-05-15
Improve bumpversion regex matching

touilleMan:
With current configuration, bumpversion will find&replace any occurrence of the version number in setup.py and my_project/init.py

This leads to trouble typically when setup.py mentions a dependency which has by hazard the same version number as our project (thus the dependency will get it version upped)

This fix precises to bumpversion a regex to match in the files in order to prevent such mistake.


audreyr:
Thanks @touilleMan!

PR #102, opened 2016-05-31
Replace Bash script test with pytest-cookies based tests

eliasdorneles:
This will make it easier to add tests using Python and increase our
coverage.

This is using [pytest-cookies](https://github.com/hackebrot/pytest-cookies)
(thanks @hackebrot !)


hackebrot:
Not at all @eliasdorneles 😸 

([cookiecutter-pytest-plugin](https://github.com/pytest-dev/cookiecutter-pytest-plugin) helps with writing plugins 🚀 )

PR #114, opened 2016-06-02
Remove distutils from setup.py Issue 111

purplediane:


eliasdorneles:
Thanks @purplediane !

PR #115, opened 2016-06-02
Added instruction to install cookiecutter using pip if haven't been i…

jhonjairoroa87:
…nstalled


audreyr:
Thanks @jhonjairoroa87!

Issue #113, opened 2016-06-02
Doublequotes in author's name generates invalid setup.py


elfsternberg:
My Github name includes my online nickname in quotes; when used with this package, the result is:

`author="Kenneth M. "Elf" Sternberg"`

... which results in an invalid setup.py.  Escapes of some sort are probably needed here.

dot2dotseurat:
Hi I'm working on this issue. :)

eliasdorneles:
fixed in https://github.com/audreyr/cookiecutter-pypackage/pull/119

Thanks @elfsternberg and @dot2dotseurat !

PR #119, opened 2016-06-02
fixed syntax error when github name has quotes

dot2dotseurat:
fixing issue https://github.com/audreyr/cookiecutter-pypackage/issues/113


eliasdorneles:
Thank you @dot2dotseurat !

PR #130, opened 2016-06-02
Remove PyPy from default Travis environments.

1337807:
A Travis build on PyPy for an unmodified cookiecutter-pypackage install will fail because it can't install the cryptography dependency. Cryptography will support the next release of PyPy but for the time being cookiecutter-pypackage can't be used on PyPy.

This is an example of a failing build: https://travis-ci.org/1337807/markovian/jobs/134879940

I initially tried using the 'pypy3' tox default environment instead of 'pypy' but that failed as well: https://travis-ci.org/1337807/markovian/jobs/134885538

I spoke with @alex an confirmed that cryptography does not support any presently released versions of PyPy. It will however support the next released version of PyPy 3 so I'm adding an issue for someone to come back and add this environment again.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27, py33, py34, py35, pypy
+envlist = py27, py33, py34, py35
 skipsdist = true
 
 [testenv]


alex:
Small clarification, we support recent versions of PyPy (everything going back to 2.6.1 in fact); it's PyPy3 that doesn't have any released versions we support.

audreyr:
Thanks @1337807 and @alex.

@skarbot says that he ran into a similar issue locally, which he resolved by upgrading pip. Might be worth trying that on Travis?

pydanny:
Rather then remove PyPy, I decided to force tox to use the latest pip in 93918bad5ddc6ed409b0b9967f2cf6c791867750. This should resolve the problems discussed in this pull request.

In case it does not, I'm keeping this open until we know it works or not.

1337807:
Sorry about the confusion @alex, I'm obviously still getting my bearings. Glad that fixed it @pydanny, I'll go find another way to get my first Python project commit. YAML changes are kind of cheating anyway. :)

1337807:
This doesn't appear to have been fixed by the pip update: https://travis-ci.org/1337807/trump_recruiter_bot/jobs/135165580

If I understand correctly these are the changes from the build log that we were hoping would fix it:

```
$ pip --version
pip 7.1.2 from /home/travis/virtualenv/python3.5.0/lib/python3.5/site-packages (python 3.5)
install
0.96s$ pip install -U tox
Collecting tox
  Downloading tox-2.3.1-py2.py3-none-any.whl (40kB)
    100% |████████████████████████████████| 40kB 3.8MB/s 
Requirement already up-to-date: py>=1.4.17 in /home/travis/virtualenv/python3.5.0/lib/python3.5/site-packages (from tox)
Collecting virtualenv>=1.11.2 (from tox)
  Downloading virtualenv-15.0.2-py2.py3-none-any.whl (1.8MB)
    100% |████████████████████████████████| 1.8MB 347kB/s 
Collecting pluggy<0.4.0,>=0.3.0 (from tox)
  Downloading pluggy-0.3.1-py2.py3-none-any.whl
Installing collected packages: virtualenv, pluggy, tox
Successfully installed pluggy-0.3.1 tox-2.3.1 virtualenv-15.0.2
9.24s$ tox -e ${TOXENV}
```

So it appears I am well beyond my depth with this issue but it seems to me that this should *just work* since cryptography is compatible with this version of PyPy. I'm having a hard time determining which version of PyPy the tox 'pypy' default is actually running. The Python version reports 3.5 but I assume that is the version that PyPy represents itself to be rather than the version of PyPy itself.

In any case I have some learning to do on Python versioning and a bit of reading on pyenv/virtualenv/pyenv-virtualenv. If I recall correctly there's something about a burrito in there as well. :)

When I figure out how to get all my local Pythons playing nicely together I should be more useful here.

PR #148, opened 2016-06-03
Fix license classifiers in setup.py

eliasdorneles:
Leftover of #135


audreyr:
Awesome, thanks @eliasdorneles!

PR #162, opened 2016-06-03
Update tox and cryptography.

1337807:
I've added RequiresIO badges (https://github.com/audreyr/cookiecutter-pypackage/pull/161) but the badge on the README is red because these dependencies are out of date. This should get us to green.


audreyr:
Great, thanks @1337807!

PR #166, opened 2016-06-04
Add sphinx doc build at tox

skarbot:
I have added sphinx build in tox.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,3 +9,10 @@ deps =
 
 commands =
     py.test
+
+[testenv:docs]
+basepython=python
+changedir=docs
+deps=sphinx
+commands=
+sphinx-build -b html -d {envtmpdir}/doctrees docs {envtmpdir}/html
--- a/tox.ini
+++ b/tox.ini
@@ -1,18 +1,17 @@
 [tox]
-envlist = py27, py33, py34, py35, pypy
+envlist = py27, py33, py34, py35, pypy, docs
 skipsdist = true
-
-[testenv]
-whitelist_externals = bash
-deps =
-    -rrequirements_dev.txt
-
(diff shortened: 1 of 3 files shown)


eliasdorneles:
Hm, what's the motivation to add Sphinx build in tox?
Looks a bit odd to me.

audreyr:
@eliasdorneles I suggested it to @skarbot yesterday and think it's a good idea. The purpose would be to ensure that the Sphinx docs build. It's possible to break the docs in such a way that they don't build, particularly when using autodoc.

I'll review this shortly once I get a free moment at the sprints 😄

eliasdorneles:
Ahh, got it, makes sense!

tox.ini
```diff
@@ -9,3 +9,10 @@ deps =
 
 commands =
     py.test
+
+[testenv:docs]
+basepython=python
+changedir=docs
+deps=sphinx
+commands=
+sphinx-build -b html -d {envtmpdir}/doctrees docs {envtmpdir}/html
```
audreyr: I believe the build is failing because line 18 needs to be indented, but I could be wrong. @skarbot, do you know?

skarbot: Sorry about that you r right @audreyr. I will fix this also it doesn't seem to work as intended. I will check and update the PR.

audreyr: Sounds good @skarbot. Hope you had a good flight back home. We're in the airport now ✈️

skarbot: I drove back to Van :) Hope you guys had a good flight. I found the mistake but now there is conflict I will fix it on my branch :)

audreyr:
Awesome work on this @skarbot, thanks!

PR #169, opened 2016-06-04
Replace perl & awk by python in make help (fixes #167 and closes #128)

### LINKED ISSUE #128: Only colorize output of `make help` if terminal has color support
Some terminals don't have support for colors, so it would be nice to detect when there is no support and avoid printing colorized output in that case.

### LINKED ISSUE #167: Remove dependency on Perl in generated Makefile

eliasdorneles:
So, there are two main motivations for this PR:

1) remove Perl & Awk dependencies (side effect: clearer code) -- see #167 
2) avoid weird output when terminal doesn't support colors -- see #128 

Point 2 is "solved" by not using colors at all.
Is this okay or am I being overly conservative?

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -10,10 +10,21 @@ except:
 webbrowser.open("file://" + pathname2url(os.path.abspath(sys.argv[1])))
 endef
 export BROWSER_PYSCRIPT
+
+define PRINT_HELP_PYSCRIPT
+import re, sys
+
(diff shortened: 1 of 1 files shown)


audreyr:
This is awesome. Thanks @eliasdorneles! Happy that you're up tonight coding like me 🌃 

Python and no colors is better than Perl and colors. In the future if we want to get fancy, we can use [Colorama](https://pypi.python.org/pypi/colorama/) or similar.

eliasdorneles:
yeaah, night coding :heart:

PR #170, opened 2016-06-04
Add makefile with command to bake project watching for changes

eliasdorneles:
When we're doing changes in cookiecutter templates, we often have to generate an re-generate the project, to check if it's actually doing what we expect.

I wanted to try making this feedback loop a bit tigher, so I came up with this. The idea is, when you're editing the template, you'd do:

    make watch

This will run `cookiecutter --no-input . --overwrite-if-exists` right away, and then every time you change something inside the template directory. I was using it while working on #169 and I quite liked it.

There is room for improvements, some things I thought:

1) it would be nice to re-run the command also when cookiecutter.json changes, or any of the hook scripts too (I couldn't figure out how to do this properly with `watchmedo`, it only watches directories)
2) it would be nice to support running cookiecutter asking for input the first time and then with `--replay` the following times

What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,4 +93,4 @@ target/
 
 # Cookiecutter
 output/
-boilerplate/
+python_boilerplate/
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,9 @@
+help:
+	@echo "bake 	generate project using defaults"
+	@echo "watch 	generate project using defaults and watch for changes"
+
+bake:
+	cookiecutter --no-input . --overwrite-if-exists
+
+watch: bake
+	watchmedo shell-command -p '*.*' -c 'make bake' -W -R -D \{{cookiecutter.project_slug}}/
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -4,3 +4,4 @@ tox==2.3.1
 cryptography==1.3.4
 cookiecutter
 pytest-cookies
+watchdog


audreyr:
I think it's a great idea! 

I tried to get it working but am having gcc errors when I try to pip install watchdog. Is there a watchdog alternative/workalike that is easier to install?

Also, you'll need to add watchdog to requirements_dev.txt.

eliasdorneles:
Hmm, yeah, one more reason to find a replacement for watchdog!
Thanks for reviewing, I'll do some digging.

eliasdorneles:
So, it looks like watchdog is the only viable cross platform solution at the moment. I looked into its alternatives ([some are mentioned here](https://github.com/gorakhargosh/watchdog/blob/master/README.rst#why-watchdog)) and none seemed a good enough solution.

The problem is that [watchdog comes with a lot of dependencies](http://pythonhosted.org/watchdog/installation.html#dependencies) (even XCode in OS X Darwin!), and so its installation can get tricky.

I also looked into [python-livereload](https://pypi.python.org/pypi/livereload) which looks cool and uses Tornado but the code there is coupled with an HTTP server (which we don't need).

I think we can leave this using watchdog for now, and replace it when we have a viable alternative (which may be something based on livereload but without the HTTP server part or perhaps in [tornado.autoreload](http://www.tornadoweb.org/en/stable/autoreload.html) which I also looked into and also didn't fit).

audreyr:
Agreed. Looks good, merging. This will be nice to have, and I'm sure I/others who are serious about Python packaging would be able to get watchdog to install with some effort.

PR #179, opened 2016-06-04
Command Line Option Added

cerickson:
I added a command line interface generation option based on click and associated tests and documentation.

The option automatically adds a project_slug entry point with noarg and --help options. Details in the console_script_setup.rst file. Cli uses click, which is used by cookiecutter itself. Click seems to have an issue with 'python setup.py develop' so I wrote up how to use 'pip install -e' in the .rst.

Testing required loading a dynamically created module and importlib is different on python 2/3.3-3.4/3.5,
so I had to do version-dependent imports/checks to run the click cli tester. Is there maybe a better way to do that? 

Let me know if you want anything changed/added/refactored. tox testing passed for all versions of python.


cookiecutter.json
```diff
@@ -10,5 +10,6 @@
 	"year": "{{ cookiecutter.release_date[:4] }}",
 	"version": "0.1.0",
 	"use_pytest": "n",
-	"use_pypi_deployment_with_travis": "y"
+	"use_pypi_deployment_with_travis": "y",
+    "create_console_script": "n"
```
audreyr: Looks like the indentation is a bit off

audreyr:
Thanks @cerickson! We'll review this shortly.

cerickson:
Thanks for reviewing! Fixed in the latest branch, the spacing error was not visible in geany but showed up correctly in emacs. ¯\_(ツ)_/¯

eliasdorneles:
Hey, nice to see this happening!

I have a some concerns about it:

I think it would be really nice if we didn't force users to use Click -- I know it is a neat framework for CLI apps, but I find it a bit too much when you just want an entrypoint script (there is all this learning curve required to using it properly and testing it too).

Note also that even though cookiecutter CLI tool relies on Click, this doesn't mean it's a free dependency for the our template: adding it to the template forces users of all the packages created with cookiecutter-pypackage to use Click when package authors accept the console script suggestion.

My main concern though, is really more about empowering people getting started and that just don't know about Click yet. I'd say that in this case, Click is just adding to the learning curve without bringing much to the table.

I'd be happy with a script based on argparse like this:

```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
"""TODO: Add script description here, will show in --help output
"""

from __future__ import print_function
from {{ cookiecutter.project_slug }} import {{ cookiecutter.project_slug }}


def run_{{ cookiecutter.project_slug }}(args):
    """
    TODO: Implement your script here
    """
    print('Input files were:', args.input_files)
    print('Ouput file will be:', args.output)

def main():
    import argparse
    parser = argparse.ArgumentParser(description=__doc__)
    # TODO: change this as you see fit:
    parser.add_argument('input_files', nargs='+',
                                           help='Input files')
    parser.add_argument('-o', '--output', default='output.txt',
                                           help='Output file')

    args = parser.parse_args()
    run_{{ cookiecutter.project_slug }}(args)


if '__main__' == __name__:
    main()
```

Another thing we could do would be to make Click optional, perhaps simply renaming the flag in this PR to be `"use_click": "n"` -- this way, we wouldn't force anyone to use it and people already familiar with Click would have that option.

It would be nice also to add a test template for the CLI tool as well.

Does this sound reasonable?
I hope I'm not being overly conservative. :)

cerickson:
Sounds reasonable to me. Either argparse or click works. Click is great, but argparse is definitely easier to test in simple user projects and in this project (just testing parse_args instead of using the click.CliRunner). I'm not super excited about all that python version + importlib stuff in the tests, although getting that working was pretty fun. Let me know which option sounds best and I'll implement it:

1. Reimplement in argparse
2. Keep click, no argparse
3. Both: Add a separate argparse section in the cli.py file and the cookiecutter.json options could be `"create_console_script": "n"/"y"/"click"`?

Cli test template is a good idea too, I'll add it after an option is chosen.

Also, moving `cli.py` back to `__main__.py`  (I started there and changed it) might be a good idea because running `python -m project_slug` and getting a console script is a cool bonus.

eliasdorneles:
I'd favor reimplementing it in argparse, but I'd love to hear @audreyr's thoughts on this.

I'd also favor keeping the code in `cli.py` because code in `__main__.py` can be hard for beginners to find. I think it can get tricky to test too.

Btw, something might be useful for the test script template is to use a context manager to inject sys.argv, like:

```
@contextmanager
def script_args(argv):
    """Capture cmdline args for testing scripts"""
    old_argv = sys.argv
    try:
        sys.argv = ['fake_script.py'] + argv
        yield sys.argv
    finally:
        sys.argv = old_argv
```

So in the tests you can write:

```
def test_when_someoption_should_do_y():
    with script_args('--someoption'):
        main()
```

audreyr:
@cerickson thank you for all your work on this, and @eliasdorneles thanks for your work reviewing!

@eliasdorneles some background: at the sprints I talked to @cerickson about this. Having been through the experience of [starting the Cookiecutter codebase with argparse](https://github.com/audreyr/cookiecutter/blob/4299379abba9a0f2121c58b8e9b6c8fe4f35a210/cookiecutter/main.py#L77) and later seeing it get factored out and replaced by Click, I thought it would be better to start with Click. But you have some great points about Click being a bit much for a simple script.

I think the best solution is to provide options for both Click and argparse. I'd suggest:

```
"command_line_interface": ["Argparse", "Click", "No command-line interface"]
```

For the purpose of this pull request, you could leave out the argparse option above.

cerickson:
Thanks for the feedback! I'll update this pull request to do the following:

1. `"command_line_interface": ["Click", "No command-line interface"]`
2. Add  command line tests using contextmanager to `test_{{ cookiecutter.project_slug }}.py`

After that I'll work on the argparse option and submit a separate pull request.

audreyr:
Sounds great @cerickson. Looking forward to it 😃

audreyr:
Great work @cerickson. It's now fully merged. 

Just a heads up that I made a few small changes, see https://github.com/audreyr/cookiecutter-pypackage/commit/a80474bc2b8ed3ea8e3c65024ddd38cae631b373:
* List of CLI options in `cookiecutter.json` rather than a single string, in preparation for allowing other option(s)
* Logic in one of the `if`s was reversed
* Broad versioning for Click to latest major version, minimizing conflicts in case the generated package and other packages using different Click versions are installed into the same env
* Error message changed since the console script already exists but is just a stub

cerickson:
Awesome, thanks for all your help! That's my first open source commit. I'll get started on the argparse feature.

PR #181, opened 2016-06-05
Fix indentation of authors in docs

purplediane:
Correct indentation of the generated toc tree in index.rst so it displays authors properly.


audreyr:
Thanks @purplediane!

PR #182, opened 2016-06-05
Add gnu gpl license to license options, add missing licenses to setup.py and tests

palmerev:


audreyr:
Great work, thanks @palmerev!

PR #188, opened 2016-06-06
Add the editorconfig file to main pypackage folder

purplediane:


audreyr:
This is great, thanks @purplediane!

PR #197, opened 2016-06-07
Fix generated read-the-docs badge

purplediane:
Read The Docs now uses .org for logins & badges. Also, they convert underscores in the repo name to dashes, sigh. They also changed the format for referencing the docs, to allow for different languages versions.


audreyr:
Thank you @purplediane! 

Do you have a link to Read the Docs' explanation about using .org for logins/badges only? It would be good to reference that here for posterity.

audreyr:
I found it: http://read-the-docs.readthedocs.io/en/latest/badges.html

PR #198, opened 2016-06-08
Prefer pip installation. Close #190

### LINKED ISSUE #190: Document which installation method is best
In `docs/installation.rst`, we should make it clear that pip installing the most recent stable release is the preferred way to install the package.

purplediane:
Is this what you were thinking about for issue 190?


audreyr:
Looks great @purplediane, thanks!

Issue #199, opened 2016-06-08
Invoking make clean fails


coleslaw481:
Date used: 6-7-2016 8:50PM PST
cookiecutter version: 1.4.0
python version: 2.7.10

### Description

Invoking make clean from a source tree generated with this cookiecutter-pypackage template results in an error.  

Here is the output:
```
$ make clean
clean-build clean-pyc clean-test
make: clean-build: No such file or directory
make: *** [clean] Error 1
```
Looks like the dependencies for the clean target were accidentally put in the command part of the target.

### What I Did

```
$ cookiecutter https://github.com/audreyr/cookiecutter-pypackage
Cloning into 'cookiecutter-pypackage'...
remote: Counting objects: 1659, done.
remote: Compressing objects: 100% (10/10), done.
remote: Total 1659 (delta 4), reused 0 (delta 0), pack-reused 1649
Receiving objects: 100% (1659/1659), 251.97 KiB | 398.00 KiB/s, done.
Resolving deltas: 100% (969/969), done.
Checking connectivity... done.
full_name [Audrey Roy Greenfeld]: bob smith
email [aroy@alum.mit.edu]: bob@bob.com
github_username [audreyr]: joe
project_name [Python Boilerplate]: foo
project_slug [foo]: 
project_short_description [Python Boilerplate contains all the boilerplate you need to create a Python package.]: 
pypi_username [joe]: 
version [0.1.0]: 
use_pytest [n]: 
use_pypi_deployment_with_travis [y]: 
Select command_line_interface:
1 - Click
2 - No command-line interface
Choose from 1, 2 [1]: 
create_author_file [y]: 
Select open_source_license:
1 - MIT license
2 - BSD license
3 - ISC license
4 - Apache Software License 2.0
5 - GNU General Public License v3
6 - Not open source
Choose from 1, 2, 3, 4, 5, 6 [1]: 
$ cd foo
$ make clean
clean-build clean-pyc clean-test
make: clean-build: No such file or directory
make: *** [clean] Error 1
```

mjscosta:
I confirm the bug, and the fix!

audreyr:
Thanks @coleslaw481 and @mjscosta. Fixed in #200 👍

PR #200, opened 2016-06-08
Fix for issue #199 make clean target failing

coleslaw481:
Hi,
  Thanks for an awesome template.  Added clean and clean-test to .PHONY target list and moved the commands for the clean target to the dependency list for the clean target. 

Hope this helps,

chris


audreyr:
Thanks for catching this and sending a fix @coleslaw481, appreciate it!

PR #203, opened 2016-06-09
Support replaying last run watching for changes

eliasdorneles:
This is a follow-up to https://github.com/audreyr/cookiecutter-pypackage/pull/170, mainly adding support for running replays.

Having this, if you want to keep baking a project using a specific set of arguments, you can:

* run it once with `cookiecutter .`, answer the prompt with the desired values
* then, run `make replay` to keep rebaking it.

You can also pass arbitrary options to cookiecutter through the `BAKE_OPTIONS` variable, e.g.: `make watch -e BAKE_OPTIONS=--replay -o other_output_dir`.

Does this look good?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,7 @@ BAKE_OPTIONS=--no-input
 help:
 	@echo "bake 	generate project using defaults"
 	@echo "watch 	generate project using defaults and watch for changes"
-	@echo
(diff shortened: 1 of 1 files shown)


audreyr:
That's awesome! Can't wait to start using this. Thanks @eliasdorneles.

PR #217, opened 2016-06-30
fix 'django' badge in readme

rochacbruno:


eliasdorneles:
oops, missed that in the review!
thanks @rochacbruno !

PR #225, opened 2016-07-28
issue208-get-windows-builds-working

JEStaubach:
appveyor.yml
    - include msys\bin to provide make
    - no command for bdist_wheel when running python setup.py so replace with bdist
Makefile
    - windows would not allow the complex python -c commands so replaced with a temp script.

Testing - Already had failed tests that are passing now. No new tests seemed to be required.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,7 +28,7 @@ environment:
       TOX_ENV: "py35"
 
 init:
-  - set PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%
+  - set PATH=%PYTHON%;%PYTHON%\Scripts;C:\MinGW\msys\1.0\bin;%PATH%
   - "git config --system http.sslcainfo \"C:\\Program Files\\Git\\mingw64\\ssl\\certs\\ca-bundle.crt\""
   - "%PYTHON%/python -V"
   - "%PYTHON%/python -c \"import struct;print(8 * struct.calcsize(\'P\'))\""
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -44,7 +44,7 @@ test_script:
   - "%PYTHON%/Scripts/tox -e %TOX_ENV%"
 
 after_test:
-  - "%PYTHON%/python setup.py bdist_wheel"
+  - "%PYTHON%/python setup.py bdist"
```
eliasdorneles: Hm, I think there is no point in distributing a binary if it's not wheel.
Do you know why it's failing to build a wheel in appveyor, even the wheel format package installs fine?

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -11,20 +11,17 @@ webbrowser.open("file://" + pathname2url(os.path.abspath(sys.argv[1])))
...
 
 help:
-	@python -c "$$PRINT_HELP_PYSCRIPT" < $(MAKEFILE_LIST)
+	echo "import re, sys" > .print_help_script.py
+	echo "for line in sys.stdin:" >> .print_help_script.py
+	echo "    match = re.match(r'^([a-zA-Z_-]+):.*?## (.*)$$', line)" >> .print_help_script.py
+	echo "    if match:" >> .print_help_script.py
+	echo "        target, help = match.groups()" >> .print_help_script.py
+	echo "        print('%-20s %s' % (target, help))" >> .print_help_script.py
+	@python .print_help_script.py < $(MAKEFILE_LIST)
```
eliasdorneles: Is this change really necessary?
Does this fail on Windows in general or only in the Appveyor build?

JEStaubach: I really hope not, there has to be a better way, I just couldn't work out how. You are right, looks like only an issue in Appveyor build. I'll take another look.

eliasdorneles:
Hey @JEStaubach, sorry for the delay, I hadn't seen these latest changes!
You did it, buddy, that was it!! Thank you so much!! :bow:

PR #228, opened 2016-07-31
Do not parameterize license text (fixes #227)

### LINKED ISSUE #227: Do no parametrize the BSD license
### Description
In https://github.com/audreyr/cookiecutter-pypackage/blob/cbcdd19a64aaa0c5bcbcc723c75628f44b13f525/%7B%7Bcookiecutter.project_slug%7D%7D/LICENSE#L27 the project name is added to a BSD license template.

I maintain https://github.com/nexB/scancode-toolkit that does detect licenses and creating multiple variants of the BSD is not a great idea for detection nor for helping downstream users.
Instead I suggest using the working of the SPDX licenses list  and not inject any name in the license templates (disclaimer I am a cofounder of SPDX)
http://spdx.org/licenses/BSD-3-Clause.html

eliasdorneles:
Avoid parameterizing BSD license text by using revised version (text from: http://spdx.org/licenses/BSD-3-Clause.html).

This is better for users too, no need to update the license text in case they decide to change the project name.

### DIFF
--- a/{{cookiecutter.project_slug}}/LICENSE
+++ b/{{cookiecutter.project_slug}}/LICENSE
@@ -24,7 +24,7 @@ are permitted provided that the following conditions are met:
   list of conditions and the following disclaimer in the documentation and/or
   other materials provided with the distribution.
 
-* Neither the name of {{ cookiecutter.project_name }} nor the names of its
+* Neither the name of the copyright holder nor the names of its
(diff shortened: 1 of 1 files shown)


eliasdorneles:
Merging this to move things forward.
Thanks for reporting @pombredanne

PR #234, opened 2016-08-17
Removing execute permissions

ParthKolekar:


eliasdorneles:
Right, that's better, thanks @ParthKolekar !

PR #240, opened 2016-08-26
Update GPL license classifier

glogiotatidis:
According to https://pypi.python.org/pypi?%3Aaction=list_classifiers the name for GPL license classifier is incomplete and pypi will not accept the package.


eliasdorneles:
Thanks @glogiotatidis !

PR #243, opened 2016-08-29
ignore .python-version file ( a pyenv configuration file)

luzfcb:
ignore .python-version file ( a pyenv configuration file)

https://github.com/yyuu/pyenv#choosing-the-python-version


eliasdorneles:
Thanks @luzfcb !

Issue #254, opened 2016-10-03
Proposed changes to test suite


mandeep:
Hi cookiecutters,

I noticed that the test suite in the project is using the unittest style tests for pytest (setup, teardown, etc.). I think these tests should be changed to use pytest fixtures and other pytest style tests. It also looks like the test function for click's CliRunner is written for both pytest and unittest, I propose just having one function in order to avoid duplication. Another thing I noticed is that the test file ends with if dunder name == dunder main. I propose that this be removed since unittest is discoverable with python -m unittest discover. One last thing, which I think can be debated, the tests directory is in the project slug's root directory. I propose moving this directory inside the package directory. Please let me know what you think and I will be happy to submit a pull request for any of the approved changes.

Proposed changes:

- Convert pytest tests from unittest style to pytest style
- Remove duplication of the cli test function (CliRunner)
- Remove if dunder name == dunder main since unittest is discoverable
- Move tests directory inside package directory

eliasdorneles:
Hm, I'd rather keep the tests directory where it is.
What's your motivation for moving them to inside the package?

The other changes sound okay to me, though I wonder if removing dunder name == dunder main could break `python setup.py test` -- that's something to check.

In any case, the best to way to move forward is to submit a PR and we can discuss it there. ;-)

mandeep:
Hi @eliasdorneles, since the tests are testing the CLI application that is in the package, it makes more sense to me to move the directory inside the package directory. I'll submit a PR and we can go from there. 

Thanks!

eliasdorneles:
Well, [both layouts are common practices](http://doc.pytest.org/en/latest/goodpractices.html#choosing-a-test-layout-import-rules) and while I understand some people might like it different, I see no point on changing the current one -- which is valid, has been tested, and users are already familiar with it.

mandeep:
Great points, I'll leave it be and focus on the other changes.

eliasdorneles:
Closed by #255

egberts:
This article details the various Python directory layouts:  https://blog.ionelmc.ro/2014/05/25/python-packaging/

PR #255, opened 2016-10-03
Removed unittest style from pytest tests and removed if __name__ == _…

mandeep:
In reference to #254 I changed the style of the pytest tests to be more inline with the typical pytest convention. I added a pytest fixture and a test function that uses the fixture as an argument. I also removed the if dunder name == dunder main from the unittest tests. With the tests directory listed in the test_suite keyword argument of setup() in setup.py, 'python setup.py test' will discover the tests automatically without the need for the dunder statement.

### DIFF
--- a/{{cookiecutter.project_slug}}/tests/test_{{cookiecutter.project_slug}}.py
+++ b/{{cookiecutter.project_slug}}/tests/test_{{cookiecutter.project_slug}}.py
@@ -26,30 +26,26 @@ from {{ cookiecutter.project_slug }} import cli
 
 
 {% if cookiecutter.use_pytest == 'y' -%}
-class Test{{ cookiecutter.project_slug|title }}(object):
+@pytest.fixture
+def some_fixture():
+    pass
 
-    @classmethod
-    def setup_class(cls):
-        pass
 
-    def test_something(self):
-        pass
+def test_something(some_fixture):
+    pass
+
 
 {%- if cookiecutter.command_line_interface|lower == 'click' %}
-    def test_command_line_interface(self):
-        runner = CliRunner()
-        result = runner.invoke(cli.main)
-        assert result.exit_code == 0
-        assert '{{ cookiecutter.project_slug }}.cli.main' in result.output
-        help_result = runner.invoke(cli.main, ['--help'])
-        assert help_result.exit_code == 0
-        assert '--help  Show this message and exit.' in help_result.output
+def test_command_line_interface(self):
+    runner = CliRunner()
+    result = runner.invoke(cli.main)
+    assert result.exit_code == 0
+    assert '{{ cookiecutter.project_slug }}.cli.main' in result.output
+    help_result = runner.invoke(cli.main, ['--help'])
+    assert help_result.exit_code == 0
+    assert '--help  Show this message and exit.' in help_result.output
 
 {%- endif %}
-
-    @classmethod
-    def teardown_class(cls):
-        pass
 {% else %}
 class Test{{ cookiecutter.project_slug|title }}(unittest.TestCase):
 
@@ -72,8 +68,4 @@ class Test{{ cookiecutter.project_slug|title }}(unittest.TestCase):
         assert '--help  Show this message and exit.' in help_result.output
 
 {%- endif %}
-
-
-if __name__ == '__main__':
-    sys.exit(unittest.main())
 {%- endif %}
--- a/{{cookiecutter.project_slug}}/tests/test_{{cookiecutter.project_slug}}.py
+++ b/{{cookiecutter.project_slug}}/tests/test_{{cookiecutter.project_slug}}.py
@@ -27,16 +27,24 @@ from {{ cookiecutter.project_slug }} import cli
 
 {% if cookiecutter.use_pytest == 'y' -%}
 @pytest.fixture
-def some_fixture():
+def response():
+    """Sample pytest fixture.
+    See more at: http://doc.pytest.org/en/latest/fixture.html
+    """
+    # import requests
+    # return requests.get('https://github.com/audreyr/cookiecutter-pypackage')
     pass
 
 
-def test_something(some_fixture):
+def test_response(response):
+    """Sample pytest test function with the pytest fixture as an argument.
+    """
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/tests/test_{{cookiecutter.project_slug}}.py
```diff
@@ -26,30 +26,26 @@
...
 {%- if cookiecutter.command_line_interface|lower == 'click' %}
-    def test_command_line_interface(self):
-        runner = CliRunner()
-        result = runner.invoke(cli.main)
-        assert result.exit_code == 0
-        assert '{{ cookiecutter.project_slug }}.cli.main' in result.output
-        help_result = runner.invoke(cli.main, ['--help'])
-        assert help_result.exit_code == 0
-        assert '--help  Show this message and exit.' in help_result.output
+def test_command_line_interface(self):
```
eliasdorneles: the `self` parameter should be removed

{{cookiecutter.project_slug}}/tests/test_{{cookiecutter.project_slug}}.py
```diff
@@ -26,30 +26,26 @@
 
 
 {% if cookiecutter.use_pytest == 'y' -%}
-class Test{{ cookiecutter.project_slug|title }}(object):
+@pytest.fixture
+def some_fixture():
+    pass
```
eliasdorneles: We could have a bit more code here to showcase how this works, for users not already familiar with pytest fixtures?

Perhaps something like:

```
@pytest.fixture
def db_transaction():
    """Sample pytest fixture
    See more at: http://doc.pytest.org/en/latest/fixture.html
    """
    # setup code, e.g. db.beginTransaction()
    yield
    # teardown code, e.g. db.rollback()
```

{{cookiecutter.project_slug}}/tests/test_{{cookiecutter.project_slug}}.py
mandeep: Good catch :smiley:

{{cookiecutter.project_slug}}/tests/test_{{cookiecutter.project_slug}}.py
mandeep: I agree, however instead of using a database transaction what do you think about creating a fixture that uses the requests library and returns a requests object? Or maybe we could include both the database transaction and the requests object to show that more than one fixture can be used?

eliasdorneles: Hm, the code shouldn't expect any dependency to be installed.
It may be best to leave just a comment.

mandeep: Thanks, I should have clarified that this would have just been a comment.

eliasdorneles: Right, much better to discuss these things after having some code. :)

So, to be honest I don't really like the fixture from the example.
Although it's illustrative of what you can do with fixtures, it doesn't represent well what one would want to do with a test.

The sample code is basically testing the fixture, and for a real test one would want to have just setup code in the fixture, and the code to be exercised in the tests itself.

Also, I'm feeling the lack of a teardown-equivalent example in pytest -- it would be nice to mirror what we have in the unittest code.

I propose:

```
@pytest.fixture
def some_fixture():
    # put your setup code to run before the test here
    yield
    # put your teardown code to run after the test run
```

Btw, I'm not completely sure if the above is the equivalent of setup / teardown -- do you happen to know if the code after the yield would run even if the test code raised an exception?

mandeep: The pytest fixture isn't really a replacement of the setup and teardown methods of unittest, I believe it's more of an improvement that changes how tests are created. If a user decided to use a pytest fixture as the setup and teardown methods in their code, they would use yield as you mentioned, and the code would continue to run in a test regardless of exception status.

I see your point about the test function testing the fixture, rather than something else independently. 

So instead of:
```
def test_response(response):
    """Sample pytest test function with the pytest fixture as an argument.
    """
    # assert response.status_code == 200
    pass
```

We could do something like:
```
def test_content(response):
    """Sample pytest test function with the pytest fixture as an argument.
    """
    # from bs4 import BeautifulSoup
    # assert 'class' in BeautifulSoup(response.content)
    pass
```

eliasdorneles: Well, okay, so what's the idiomatic way of doing setup/teardown w/ pytest?

Do we leave that to users figure out? Do we assume users that choose to use pytest already know it?

mandeep: The documentation lists this example for setup/teardown:

```
import smtplib
import pytest

@pytest.fixture(scope="module")
def smtp(request):
    smtp = smtplib.SMTP("smtp.gmail.com")
    yield smtp  # provide the fixture value
    print("teardown smtp")
    smtp.close()
```

I'm not really sure if we should provide an example of this to the user, or if we should let them read the documentation to figure it out. On IRC, the biggest problem I see with users new to pytest is that they are unable to wrap their head around being able to define functions as fixtures that are then passed as arguments to the test functions. From what I've seen, once they see how fixtures are used, they typically figure out how to do the rest.

eliasdorneles:
Thanks @mandeep !

mandeep:
@eliasdorneles Thank you! I'm okay with you closing #254 if need be.

mandeep:
@eliasdorneles I just noticed that line 42 in the tests file states:

```
# assert 'GitHub' in BeautifulSoup(r.content).title.string
```

when it should be:

```
 # assert 'GitHub' in BeautifulSoup(response.content).title.string
```

My apologies for this typo. Should I submit a new pull request with the fix?

eliasdorneles:
Oh, I missed it in the review.
Yea, feel free to send a new PR, thanks.

PR #262, opened 2016-10-22
On pypi.org, keywords are a list of strings

benjaoming:
Example:
https://pypi.org/project/wiki/

Code:
https://github.com/django-wiki/django-wiki/commit/688999ebce5cc2ddeb65c9383fd375c2f7b9dc26

### DIFF
--- a/{{cookiecutter.project_slug}}/setup.py
+++ b/{{cookiecutter.project_slug}}/setup.py
@@ -54,7 +54,7 @@ setup(
     license="{{ cookiecutter.open_source_license }}",
 {%- endif %}
     zip_safe=False,
-    keywords='{{ cookiecutter.project_slug }}',
+    keywords=['{{ cookiecutter.project_slug }}',],
     classifiers=[
         'Development Status :: 2 - Pre-Alpha',
         'Intended Audience :: Developers',


eliasdorneles:
Has this changed recently?

From a quick google research, I only find it as a space separated string, including [the PyPA sample project](https://github.com/pypa/sampleproject/blob/master/setup.py#L68) and [setuptools basic example](http://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=keywords#basic-use).

benjaoming:
It's a bit confusing... but after this [1] change, keywords need to be either comma/semicolon-separated or as a list. I don't think docs are updated correctly. I can try verifying the actual state of things.. if what has happens is what I think it is, then PyPi.org development should assure with their changes that they're 100% docs-driven.

[1] https://github.com/pypa/warehouse/pull/975

eliasdorneles:
From a quick look at that code, it looks like space-separated is still supported: https://github.com/pypa/warehouse/pull/975/files#diff-387b0e7ca289cf745391093d2e5f6f1dR108

And it should, it would be silly to require every package to change their setup.py :)

benjaoming:
Okay, I had the issue not more than a couple of weeks ago on pypy.org so I figured the PR from february wasn't it.. but the issue *is* fixed now, verified it... spaces do work. The issue was that pypi.org would display "k, e, y, w, o, r, d, 1"

Thanks for following up @eliasdorneles 

Closing this, since it's not decided to formalize the list version in the docs (it does work on the old pypi.python.org, too).

eliasdorneles:
Gotcha!
Thanks for digging it, anyway! :+1:

PR #263, opened 2016-10-25
Remove auto-generated comment :)

benjaoming:
I guess the generated docs files are quite modified compared to what sphinx autogenerates, so we can remove this :)


benjaoming:
Apparently I end all sentences with smileys right now :)

eliasdorneles:
Right, makes sense!

PR #267, opened 2016-11-07
Invoke py.test correctly with coverage run

benjaoming:
If I run `make coverage` (Python 3.5), I get:

``` 
$> make coverage 
coverage run --source my_project py.test
No file to run: 'py.test'
Makefile:64: recipe for target 'coverage' failed
make: *** [coverage] Error 1
```

Correct invocation is to use the -m switch.

http://stackoverflow.com/a/16524426/405682


eliasdorneles:
Thanks @benjaoming !

PR #284, opened 2016-12-14
ENH: cli:main: sys.exit(zero_returncode)

westurner:


pydanny:
Can you point to where this practice is documented within Click?

westurner:
I suppose this is redundant but otherwise good practice.

http://click.pocoo.org/5/exceptions/#where-are-errors-handled

- https://github.com/pallets/click/issues/270
  - https://github.com/pallets/click/pull/765
- https://github.com/pallets/click/issues/747

westurner:
(as long as it's correct) it's a best practice to explicitly return a zero returncode on non-error so that other commands can be logically chained:

```bash
cookiecutter <args> && echo 'Done'
```

pydanny:
Wow, thanks for instructing me!

PR #294, opened 2017-01-23
lgiordani fork added

lgiordani:
I added my fork that uses the punch version updater instead of bumpversion


lgiordani:
Thank you!

PR #299, opened 2017-02-14
Use find_packages instead

jonathan-s:
I recently created a project and using find_packages instead of manually defining where the packages are helped me get this working. With manually defining the packages it's easy to get in a situation where you get import errors (which I did). So it's really preferable to use `find_packages`

### DIFF
--- a/{{cookiecutter.project_slug}}/setup.py
+++ b/{{cookiecutter.project_slug}}/setup.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-from setuptools import setup
+from setuptools import setup, find_packages
 
 with open('README.rst') as readme_file:
     readme = readme_file.read()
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_slug}}/setup.py
```diff
@@ -36,11 +36,7 @@
     author="{{ cookiecutter.full_name.replace('\"', '\\\"') }}",
     author_email='{{ cookiecutter.email }}',
     url='https://github.com/{{ cookiecutter.github_username }}/{{ cookiecutter.project_slug }}',
-    packages=[
-        '{{ cookiecutter.project_slug }}',
-    ],
-    package_dir={'{{ cookiecutter.project_slug }}':
-                 '{{ cookiecutter.project_slug }}'},
+    packages=find_packages(),
```
eliasdorneles: One problem of using `find_packages` is that it might find packages that you don't really want to include in the final distribution.

In this case, for example, you're including all the tests in the distribution as a `tests` module, which may also lead to import errors:

```
>>> from setuptools import find_packages
>>> find_packages()
['python_boilerplate', 'tests']
```

You could fix it by using `find_packages(exclude=['tests'])`, but it could still potentially add other unwanted stuff (e.g. code generated automatically by other tools).

So the design decision to be made here is: is it better to 1) expect people to specify explicitly which modules to include (and adding the default) or 2) add everything that can be added (excluding tests), which can potentially add clutter to people's packages and therefore to the systems of people that end up installing that package?

Maybe a good compromise is to use `find_packages(include=['{{ cookiecutter.project_slug }}'])`, which will behave the same as it does now, but at least people that want to use it differently can just change the arguments to `find_packages()`.

jonathan-s: I think that sounds like a fair compromise. I'll can make changes that accommodates that.

eliasdorneles:
Sorry for the delay, thanks @jonathan-s  !

PR #308, opened 2017-03-27
Updates README.rst with svg badge

sobolevn:


eliasdorneles:
Hi @sobolevn -- could you explain why this is better?
It looks better, just asking in case there is some other reason than looks. :)

sobolevn:
`svg` looks way better on retina displays. no other reason.

eliasdorneles:
Ah okay, thanks, looks good! :+1:

PR #314, opened 2017-04-26
Use the correct number of `=` to make a title.

Natim:
With regards to the size of the `project_name` we should not add the same number of `=`


Natim:
Refs https://github.com/pytest-dev/cookiecutter-pytest-plugin/blob/template-macro-in-readme/pytest-%7B%7Bcookiecutter.plugin_name%7D%7D/macros/rst

eliasdorneles:
Nice work, thanks @Natim !

PR #316, opened 2017-05-01
kragniz/cookiecutter-pypackage-minimal - relevant?

nealmcb:


pydanny:
I don't think cookiecutter-pypackage-minimal is maintained anymore. Please reopen if this is not the case.

PR #317, opened 2017-05-02
Change the way TravisCI is used.

mpasternak:
Instead of using single python version for all builds on Travis and then using tox to setup different interpreters in different environments, setup TravisCI integration in a way such, that it uses its own mechanisms to handle Python installation and then runs tox with only a single environment (single python version) inside the build.

This way you won’t get errors about missing Python interpreter from TravisCI, that you had chances to get in the previous configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,4 @@ target/
 # Cookiecutter
 output/
 python_boilerplate/
+\#*\#
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,17 @@
 language: python
 
-python: 3.5
-env:
-  - TOXENV=py27
-  - TOXENV=py33
-  - TOXENV=py34
-  - TOXENV=py35
+python: 
+  - 2.7
+  - 3.3
+  - 3.4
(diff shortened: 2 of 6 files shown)


browniebroke:
I wanted to use this cookie cutter to generate a new package and found this PR. I'm interested in it because it's adding Python 3.6 to Travis 👍 
I think it would make sense to update the README to states that and also the classifiers in `setup.py`, right?

mpasternak:
@browniebroke , if you want to submit some pull requests, I'm willing to add them to my branch. I plan to add some more stuff to my branch, I already added "github_project_slug" because I'd rather use dashes than underscores and I want to have it different; I plan adding templates for appveryor.com (Windows) builders... I also thought a bit about adding MacOS X support to TravisCI files, so there is still much stuff to do.

pydanny:
Over time we've organically moved to something that isn't far off from this PR. I may cherry-pick portions of @mpasternak's effort into the main branch, his `Makefile` enhancements in particular look very useful.

PR #318, opened 2017-05-06
Support hyphens in project name. Fixes #250

### LINKED ISSUE #250: Project slug suggestion should replace hyphen (-) for underscore (_)
- Date you used Cookiecutter PyPackage: 2016-10-26
- Cookiecutter version used, if any: 1.4.0
- Python version, if any: 3.5.2
- Operating System: ArchLinux
### Description

When you create a project, the project_name can contains hyphen, but in the next step, the project_slug can't. The project_slug suggestion is the project_name and you can put the same value, but  then will fail because "The project slug (foo-bar) is not a valid Python module name. Please do not use a - and use _ instead".
So i think the default value should be project_name.replace("-", "_")

browniebroke:
Hyphens in project names were considered invalid as per the existing test.

It should be possible to support them by replacing hyphens by underscores when generating the default slug, as suggested in #250.

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -3,7 +3,7 @@
   "email": "aroy@alum.mit.edu",
   "github_username": "audreyr",
   "project_name": "Python Boilerplate",
-  "project_slug": "{{ cookiecutter.project_name.lower().replace(' ', '_') }}",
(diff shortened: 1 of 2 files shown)


eliasdorneles:
Looks good, thanks @browniebroke !

browniebroke:
✨

PR #329, opened 2017-06-08
Integrate pytest into setup.py

astralblue:
That is, make 'setup.py test' use pytest if pytest is enabled.

### DIFF
--- a/{{cookiecutter.project_slug}}/setup.cfg
+++ b/{{cookiecutter.project_slug}}/setup.cfg
@@ -16,3 +16,19 @@ universal = 1
 
 [flake8]
 exclude = docs
+
+[aliases]
+{%- if cookiecutter.use_pytest == 'y' %}
+test = pytest
+{%- endif %}
+# Define setup.py command aliases here
+{%- if cookiecutter.use_pytest == 'y' %}
+
+[pytest]
+# Set permanent options for the "pytest" setup command here
+#index-url = http://my-pypi.example.com/
+
+[tool:pytest]
+#testpaths = tests
+# Set permanent options for py.test itself here
+{%- endif %}
--- a/{{cookiecutter.project_slug}}/setup.cfg
+++ b/{{cookiecutter.project_slug}}/setup.cfg
@@ -22,13 +22,3 @@ exclude = docs
 test = pytest
 {%- endif %}
 # Define setup.py command aliases here
-{%- if cookiecutter.use_pytest == 'y' %}
-
-[pytest]
-# Set permanent options for the "pytest" setup command here
-#index-url = http://my-pypi.example.com/
-
-[tool:pytest]
-#testpaths = tests
-# Set permanent options for py.test itself here
-{%- endif %}
--- a/{{cookiecutter.project_slug}}/setup.py
+++ b/{{cookiecutter.project_slug}}/setup.py
@@ -18,14 +18,14 @@ requirements = [
 
 setup_requirements = [
 {%- if cookiecutter.use_pytest == 'y' %}
-    pytest-runner
+    'pytest-runner',
 {%- endif %}
     # TODO({{ cookiecutter.github_username }}): put setup requirements (distutils extensions, etc.) here
 ]
 
 test_requirements = [
 {%- if cookiecutter.use_pytest == 'y' %}
-    pytest
+    'pytest',
 {%- endif %}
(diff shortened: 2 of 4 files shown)


{{cookiecutter.project_slug}}/setup.py
```diff
@@ -16,7 +16,17 @@
     # TODO: put package requirements here
 ]
 
+setup_requirements = [
+{%- if cookiecutter.use_pytest == 'y' %}
+    pytest-runner
```
eliasdorneles: This doesn't look right, entries in this list need to be Python strings.

Can you please add a test for this?

I think it's just a matter of adding [a line like this](https://github.com/audreyr/cookiecutter-pypackage/blob/1d97c602eed3db2c6cfeb71882c94fc00845ef5f/tests/test_bake_project.py#L93) to [`test_using_pytest()` here](https://github.com/audreyr/cookiecutter-pypackage/blob/1d97c602eed3db2c6cfeb71882c94fc00845ef5f/tests/test_bake_project.py#L181-L186).

astralblue: Oops.  :-p  Will fix (and add test).

astralblue: By the way, isn't that ``run_inside_dir(...) == 0`` line supposed to be prefixed by an ``assert``?  Simple comparisons without assertion won't make tests fail even if they fail….

eliasdorneles: Oh yeah, that comparison isn't accomplishing anything.
Frankly adding an assertion won't do much either, because `subprocess.check_output` will raise an exception if the command return non-exit status.

You can leave out the `== 0` or add an assert, I don't mind either. =)

astralblue:
Fixed and added tests.

I will address the ``run_inside_dir(...) == 0`` dichotomy in a separate PR, in the spirit of atomic PRs (which I concur are awesome).  :)

{{cookiecutter.project_slug}}/setup.cfg
```diff
@@ -16,3 +16,19 @@ universal = 1
...
+
+[aliases]
+{%- if cookiecutter.use_pytest == 'y' %}
+test = pytest
+{%- endif %}
+# Define setup.py command aliases here
+{%- if cookiecutter.use_pytest == 'y' %}
+
+[pytest]
+# Set permanent options for the "pytest" setup command here
```
eliasdorneles: So, I was just trying this in a clean virtualenv and got the following warning:

```
============================================== warnings summary ===============================================
None
  [pytest] section in setup.cfg files is deprecated, use [tool:pytest] instead.

-- Docs: http://doc.pytest.org/en/latest/warnings.html
==================================== 2 passed, 1 warnings in 0.10 seconds =====================================
```

Does it make sense to have these two sections?
Sounds like we should use only `tools:pytest`, no?

astralblue: ``[pytest]`` is for the “pytest” distutils command, served by the pytest-runner setuptools extension.  pytest itself historically used the same section, but last year it started using ``[tools:pytest]`` instead – after the pytest-runner author complained in pytest-dev/pytest#567 that pytest itself, not being a distutils command, should not occupy a section whose name is not colon-prefixed.

astralblue: IMO that pytest warning is a little too aggressive.  A better approach would be to raise the warning only when the section actually contains a known pytest configuration directive.

eliasdorneles: I see, thank you very much for the explanation and pointers! :+1: 
So, considering that 1) both sections are empty and that 2) template users can add a `pytest.ini` if they want to add pytest options, what if we simply leave out these and let the template users decide what to do?

astralblue: That works for me.  Will update.

eliasdorneles:
Hi @astralblue, this looks good to me, but the AppVeyor build failed.
I think it's a temporarily failure, but am not sure, and sadly I don't have a way of restarting it.
Can you add a new commit just to trigger a new build, so that we verify this hypothesis?
Thank you!

astralblue:
Voila!

eliasdorneles:
Thank you @astralblue ! :tada:

PR #330, opened 2017-06-08
Fix indentation of coverage target in Makefile

astralblue:


{{cookiecutter.project_slug}}/Makefile
```diff
@@ -61,14 +61,14 @@ test-all: ## run tests on every Python version with tox
...
 coverage: ## check code coverage quickly with the default Python
-	{% if cookiecutter.use_pytest == 'y' -%}
-		coverage run --source {{ cookiecutter.project_slug }} -m pytest
-	{% else %}
-		coverage run --source {{ cookiecutter.project_slug }} setup.py test
-	{% endif %}
-		coverage report -m
-		coverage html
-		$(BROWSER) htmlcov/index.html
+{% if cookiecutter.use_pytest == 'y' -%}
```
eliasdorneles: Hello @astralblue, nice catch,  thanks for the PR!

This is adding some extra whitespace though, it will be better to move the `-` that strips whitespace to the beginning of `if` line, like so:

    {%- if cookiecutter.use_pytest == 'y' %}

And do the same for the `else` and `endif`.

With that change, this will be good to merge. :+1: 

If you want to be extra nice, you can do it for the `test:` target which seems to have the same problem. :)

Thank you!

astralblue: Thanks for the review!  Fixed the issue.

eliasdorneles:
Thanks @astralblue !

PR #331, opened 2017-06-08
Keep trailing whitespace in diff files

astralblue:
Without this, editorconfig-vim strips trailing whitespace in context and
old lines in unified diff (used frequently in the git add --patch
workflow), upon saving; the resulting diff fails to apply.


eliasdorneles:
Makes sense, thanks @astralblue !

PR #333, opened 2017-06-10
Use Tox-Travis for environment selection

astralblue:
Tox-Travis supports mapping each Travis factor (a value of os, language,
python, environment variables, etc.) onto zero or more Tox environment
name factors, so as to fully take advantage of generative listing of
environments in Tox 1.8+.  It eases testing with a matrix of dependency
version combinations.

Examples are in https://tox-travis.readthedocs.io/en/stable/toxenv.html

### DIFF
--- a/{{cookiecutter.project_slug}}/.travis.yml
+++ b/{{cookiecutter.project_slug}}/.travis.yml
@@ -2,21 +2,18 @@
 # This file will be regenerated if you run travis_pypi_setup.py
 
 language: python
-python: 3.5
-
-env:
-  - TOXENV=py35
-  - TOXENV=py34
-  - TOXENV=py33
-  - TOXENV=py27
-  - TOXENV=py26
-  - TOXENV=pypy
+python:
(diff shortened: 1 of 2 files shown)


eliasdorneles:
So, I suppose the broken build illustrates the advantage of the existing setup: it's compatible with Appveyor (and other non-Travis build systems). =)

astralblue:
Not sure if that failure is directly caused by this change, because the build history shows that [another adjacent build](https://ci.appveyor.com/project/audreyr/cookiecutter-pypackage/build/1.0.328) triggered by pyup.io failed for the similar reason (OpenSSL compilation failure).  Not sure why that is happening though….

astralblue:
Voila! (2)

eliasdorneles:
Oh, I had misunderstood this PR before, this is actually pretty cool!
It enables one to set up a build matrix combining other env variables with the different python versions. :+1: 
Thanks!

PR #336, opened 2017-06-11
Remove package directory from PYTHONPATH

astralblue:
Without this, if we had <project_slug>/xyz.py, it would be importable
both as "<project_slug>.xyz" (correct) and as "xyz" (incorrect).


astralblue:
By the way, sorry for the whole salvo of PRs… these are things that I had to fix in 5+ projects created from this cookiecutter. :p

eliasdorneles:
Hey @astralblue -- no worries, your contributions are very much appreciated! <3
Some might take some time to review because we're all busy with other things, I look at them when I have some little time to spare.

PR #339, opened 2017-06-22
Fixing bug with authors and history headings running together in the docs

fhightower:
When using both an authors and history files, the entries for both of these files in the docs run together. It looks like:
`authorshistory`
rather than:
```
authors
history
```

This PR fixes the problem.


eliasdorneles:
Hi @fhightower , thanks for reporting!
Hm, we had this problem earlier and #323 fixed it for me.
Can you reproduce this problem with the current code from master?
Also, which version of Sphinx and Jinja are you using?

fhightower:
Ah, that's my fault. I had an older version of jinja. It is working properly now.

Thanks @eliasdorneles .

PR #350, opened 2017-07-21
Proposal for SUPPORT.md

benjaoming:
I would like to propose this draft of SUPPORT.md

I did NOT insert an auto-generated IRC channel name on purpose. I don't think this info should be added without the maintainer knowing, it's okay to add later.

In any case, people probably have different modes for offering support - but I think the bulk of maintainers do not want to be the target of support enquiries but prefer that all such are directed to community channels.


pydanny:
Being someone who is inundated with requests for help, I've found not explaining support channels has been a good thing. Especially since the chance of anyone actually paying for support is about as likely as improved gun control in the USA.

However, we're considering becoming more aggressive with offering "paid support" plans for Cookiecutter PyPackage and Cookiecutter and other things we maintain. Maybe we can make enough to buy dinner in a fast food restaurant once a month. In which case, what do you think of linking to that in SUPPORT.rst documents?

benjaoming:
@pydanny my main objective here isn't to suggest what the Cookiecutter project should do :) but I really hope you manage to balance your work load in a way that also covers your living!!

The intention was to suggest a default for Github projects in general which are built with `cookiecutter-pypackage`.

I personally prefer referring support inquiries to other channels if they are a burden. It's easier for community to provide support to others than to provide contributions such as issue reports and pull requests. But if you like doing support, perhaps paid support is an interesting path to explore. That would also make the cost of support more evident to people demanding or expecting free support.

There's also Patreon btw.

benjaoming:
I would also add this: I see support inquiries usually as a lack of documentation *or* because someone didn't read it (in which case it's a bit easier to give support). Usually people who receive support do not amend the docs, and I often end up responding without amending them myself.

The problem here is: Of course the support instructions will tell people to RTD. But since they are not the types who read SUPPORT.rst, then they also are not the types who RTD :)

And that can be a pain and will always be, no matter what we want to write in SUPPORT.rst.

PR #379, opened 2017-10-24
Use Tox-Travis for environment selection (redux)

astralblue:
This is PR #333 applied to cookiecutter-pypackage itself.

Also retire Python 3.3 (EOL; no longer available on Travis-CI) and add Python 3.6.


astralblue:
This PR fixes Travis-CI builds (which had been broken since Python 3.3 EOL).

astralblue:
Appveyor builds for Python 3.6 failed due to an OpenSSL compilation error (the same transient error we had seen before).  All the other builds are fine.

pydanny:
Thanks for this contribution but it's been fixed in another PR.

PR #392, opened 2018-01-04
Fix build issue on debian and ubuntu

askz:
This was tested in debian 9 (stretch)


pydanny:
Fixed in a separate pull request

PR #394, opened 2018-01-07
Fix failing Travis build

ErwinJanssen:
The Travis build of projects generated with this cookiecutter fail out of the box. The Travis build for the project itself is also failing. This pull request attempts to resolve this by:

- Updating the test setup of the project to match the setup of the generated project. The generated project seems to have a more modern approach, and keeping this uniform will improve maintainability.
- Update all dependencies to the most recent versions.
- Remove support for Python 2.6 (no longer supported by the Python core team) and Python 3.3 (not supported by Pytest)

Additionally, support for Python 3.6 was added, since this is the most recent version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ python:
   - 3.4
   - 3.3
   - 2.7
-  - 2.6
 
 # command to install dependencies, e.g. pip install -r requirements.txt --use-mirrors
 install: pip install -U tox-travis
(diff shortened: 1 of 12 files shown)


hugovk:
👍 

> Remove support for Python 2.6 (no longer supported by the Python core team) and Python 3.3 (not supported by Pytest)

Yep, both Python 2.6 and 3.3 are no longer supported by the Python core team nor Pytest.

You could also update this to `(3, 4)` in test_bake_project.py:
```python
    elif sys.version_info >= (3, 3):
```

(Then when 3.4 is ditched, the whole `elif` block can be removed.)

PR #408, opened 2018-02-14
Add support for optional tools (VSCode, YAPF)

tmattio:
This PR brings 4 changes:

- Support for VSCode
- Support for YAPF
- Remove some files that are only relevant when the project is open source

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -8,6 +8,7 @@
   "pypi_username": "{{ cookiecutter.github_username }}",
   "version": "0.1.0",
   "use_pytest": "n",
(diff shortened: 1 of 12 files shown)


pydanny:
Thank you so much for this contribution. However, to keep complexity and scope down, this particular template stays agnostic in terms of IDE or text editor (hence why we have `.editorconfig` but not pycharm config). Also, as we already have Flake8, YAPF is unnecessary. 

However, if you want to create a maintained fork with your enhancements, we encourage you to do so. In fact, once you do we'll even link to the in the README section called 'Similar Cookiecutter Templates'.

PR #426, opened 2018-05-13
Use pytest instead of py.test per upstream recommendation, #dropthedot

scop:
http://blog.pytest.org/2016/whats-new-in-pytest-30/
https://twitter.com/hashtag/dropthedot


scop:
Oops, didn't remember I had already filed this. Anyway prefer new up to date one in #429

PR #431, opened 2018-07-19
BLD: setup.py: add Python 3.7 to PyPI classifiers list

westurner:


westurner:
- [ ] Add 3.7 to tox.ini
  https://github.com/audreyr/cookiecutter-pypackage/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/tox.ini

What else?

pydanny:
As there shouldn't be any reason why AppVeyor is failing, I'm just going to merge this in. I'll look into why AppVeyor is failing tonight.

westurner:
I should've added py37 to tox.ini first.


Is there also a reference to 3.6 support in the README?

PR #449, opened 2018-12-18
Fix missing assert in some tests

javiersanp:
I suppose that all the run_inside_dir(..., ...) == 0 was meant to have an assert


PsiACE:
Thank you for your PR. Adding `assert` does not bring substantial benefits. Please let me know if this is necessary.

AndreMiras:
Either the `== 0` isn't necessary or the `assert` is very necessary, right?
What do you think @PsiACE ?
I'm would be happy to rebase this PR and fix one or the other

AndreMiras:
Why was this closed? You have that code `run_inside_dir('pytest', str(result.project)) == 0`, why do you need the `== 0` if you don't make it fail when it's not?

PR #450, opened 2019-01-04
update dev requirements

umeboshi2:
Just an update to the latest version.


PsiACE:
We use the pyup to manage the dependency upgrade, and the PR has been a long time, some new changes have occurred, I will turn it off.

You're welcome to contribute in other ways.

PR #459, opened 2019-03-01
Modify find_packages usage to find slug.* modules

### LINKED ISSUE #423: Consider modifying usage of find_packages
Currently, [find_packages is used](https://github.com/audreyr/cookiecutter-pypackage/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/setup.py#L61) as follows:

    packages=find_packages(include=['{{ cookiecutter.project_slug }}']), 

If the user creates a package `foo` and then creates a subpackage `foo.bar` (i.e. `bar` is a directory with an `__init__.py`), then the subpackage needs to be added explicitly to the packages list in `setup.py`.

I think that this can lead to unexpected consequences. For example, suppose the user forgets to add `foo.bar` to the package list. All tests will pass locally if they run `tox` or `pytest` from the project root directory because `foo.bar` is on the system path. All tests will pass on CI because the tests are executed from the project directory again. But if an end-user then pip installs the package, `import foo.bar` will result in an import error because the `bar` package will not have been installed to `site-packages` per `setup.py`. (This just happened to me, and it was impossible to debug until I finally `ls`ed the `site-packages` directory of the environment I installed my own package to.)

I suggest to consider a slightly different paradigm which is used by Pandas [for example](https://github.com/pandas-dev/pandas/blob/master/setup.py#L735).

    packages=find_packages(include=['{{ cookiecutter.project_slug }}', '{{ cookiecutter.project_slug }}.*']), 

The usual downside to this usage would be that a `tests` subpackage of `foo` would be installed, but that is not an issue in this pypackage organization.

Thoughts?

micahjsmith:
Modify find_packages use in setup.py to find packages named `slug` or `slug.*`, where slug is the `project_slug`.

You can see that currently, only the root module is detected by find_packages. With the proposed fix, submodules are also detected.

Here is a MWE:
```
$ cookiecutter --no-input gh:audreyr/cookiecutter-pypackage
$ cd python_boilerplate
$ mkdir python_boilerplate/{foo, bar}
$ touch python_boilerplate/{foo, bar}/__init__.py
$ python -c 'import setuptools;print(setuptools.find_packages(include=["python_boilerplate"]))'
['python_boilerplate']
$ python -c 'import setuptools;print(setuptools.find_packages(include=["python_boilerplate", "python_boilerplate.*"]))'
['python_boilerplate', 'python_boilerplate.foo', 'python_boilerplate.bar']
```

Would close #423

### DIFF
--- a/{{cookiecutter.project_slug}}/setup.py
+++ b/{{cookiecutter.project_slug}}/setup.py
@@ -59,7 +59,7 @@ setup(
     include_package_data=True,
     keywords='{{ cookiecutter.project_slug }}',
     name='{{ cookiecutter.project_slug }}',
-    packages=find_packages(include=['{{ cookiecutter.project_slug }}']),
+    packages=find_packages(include=['{{ cookiecutter.project_slug }}', '{{ cookiecutter.project_slug }}.*']),
     setup_requires=setup_requirements,
     test_suite='tests',
     tests_require=test_requirements,


PsiACE:
Well done.

PR #471, opened 2019-04-07
PyPI link update

gokceneraslan:


PsiACE:
Thanks to your PR, we have a lot of links about PYPI that need to be updated. Can you continue to help?

PR #476, opened 2019-04-23
Add Argparse CLI option, unit test and documentation

jamwil:
This request creates an additional option to use argparse for console scripts. I realized rather late into the process that argparse was only added in Python 3.2, so this isn't likely to meet the criteria of Python 2.7 support. Figured I would submit it for conversation's sake and would very much appreciate any feedback.


PsiACE:
Nice to see this PR, I will test it further and then merge.

PR #481, opened 2019-05-12
Move cli.py to __main__.py to allow for `python -m package`

zdog234:
Hello! I love this template. While using it at a pycon tutorial, I noticed that it doesn't have a `__main__.py`, so it's not possible to run `python -m package_name`. I thought I'd submit a PR to make that possible.

Let me know what i'm missing, or if this is totally off-base. And thanks again for the awesome template :)


PsiACE:
Nice to see you working on it. I'll check on it later.

PsiACE:
I thought we could do something like flask, just add one `__main__.py` file.
Please let me know if there are any intuitive benefits to removing the `cli.py`.

zdog234:
Hmm. I think my intuition was that it's simpler to just have `__main__.py`, but I think I can see a few more benefits for the approach that flask uses.

1. It might make sense to have some code in `cli.py` that you'd actually want to import. (i.e. `mypackage.cli`). (Now that I think about it, I've actually ran into an issue with this in the past, where I had some code in my `__main__.py` file that I had to pull out into a separate module so that I could import it somewhere else. If I had put all of that into a `cli.py` file from the beginning, I never would've had to do that.
2. `cli.py` just seems like a more descriptive name than `__main__.py` -- I still get confused by how dunder things work sometimes.

Also, it doesn't seem like that big of a deal to just add one more file, especially considering how many files tend to end up in a python package project 

Would you like me to refactor this PR to use an approach more like flask's?

PsiACE:
Yeah, I think that's better. I'd be happy to merge it if you refactor it.

PR #501, opened 2019-07-17
fixes: if Click is choosen it doesn't show up in the requirements_dev.txt

s3h10r:
Hi, there's a tiny bug in the requirements_dev.txt Template which prevents correct representation of the selected command_line_interface option. If 'Click' is choosen it shows up in the Project's `setup.py` but not in its `requirements_dev.txt`. This patch fixes this tiny problem. Greetings & thanks for your awesome project, Sven

### DIFF
--- a/{{cookiecutter.project_slug}}/requirements_dev.txt
+++ b/{{cookiecutter.project_slug}}/requirements_dev.txt
@@ -7,6 +7,9 @@ tox==3.5.2
 coverage==4.5.1
 Sphinx==1.8.1
 twine==1.12.1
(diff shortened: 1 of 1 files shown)


PsiACE:
You can try `python setup.py develop`.

Do we really need to add it?

s3h10r:
Right of course, i just stumbled over the lines in upstream 
```
{% if cookiecutter.command_line_interface == 'y' -%}
click==6.7{% endif %}
```
which seems just not work. So i thought it maybe is a good idea to correct this - or throw the lines out. Was a bit confusing imho.

PsiACE:
You are right, I will check it later and merge.

Issue #445, opened 2018-12-02
Support pipenv as package manager


Serkan-devel:
[pipenv](https://pipenv.readthedocs.io/en/latest/) is a package manager, which makes dependency tracking and updating easier for each individual project.

jberends:
indeed -- agreed. We need to move away from reqs.txt and into pipenv to create a better-world :tm:

PsiACE:
That's a good idea. We'll add pipenv in the future, and we'll also consider poetry.

moodule:
well change is coming your way :steam_locomotive: 

I just made a PR which should give pipenv support out of the box.

hopefully I didn't blow this and it's at least a first step :D

now:
- "make init" creates a pipenv and installs the deps in it
- all the other rules are executed inside this env (test, tox etc)

Zethson:
@moodule would you be interested in adding your fixes to the cookietemple cli-python template?

https://github.com/cookiejar/cookietemple/

The cli-python template originates from this template here and could certainly use your fixes.

audreyfeldroy:
The template now uses uv for dependency management, which handles this use case.

PR #519, opened 2019-09-16
Bundle all the dependencies in a pipenv

moodule:
Add make rules to encapsulate all the operation in a pipenv.

Makes sharing / developping the generated package easier.


moodule:
this should -at least partially- solve #445

PR #520, opened 2019-09-16
Add a code of conduct

apehex:
Might come in handy for (wannabe) ambitious / popular projects :)


PsiACE:
I'm sorry to reply you so late. I like the idea, but it should be an option, not a necessity. 
I'd be happy to merge it if you'd like to modify it.

PR #523, opened 2019-09-16
Document how to add a new test for new features / bug fixes

moodule:
The test should do 2 things:
- show what the code lacked before the changes
- prove that everything works after the changes


moodule:
I meant: this is why I felt having a few guidelines for creating new tests might be useful.

I'm not changing the bake tests on the cookiecutter, just documenting the tests on the generated package.

PR #524, opened 2019-09-16
Rename dependency file to follow pythonic standard (?)

moodule:
May-be I'm missing something here, but this "requirements_DEV.txt" hurts my eyes and I don't get the use of that "_dev" suffix?!

Plus, this file is somewhat obsolete with pipenv, no?

Also sorted the deps cause it's easier to add new ones that way.


PsiACE:
This suffix is used to identify the dependencies that are applicable to the development phase.

Semantically, it seems to be acceptable.

moodule:
well yeah but isn't it the point of the "requirements.txt" file from the start?

so to me it kinda feels like being handed a menu that says "menu of the recipes served here" in a restaurant :D

PsiACE:
Since it includes Click, I think your opinion is appropriate.

I'll consider merging this. But it's a lower priority.

moodule:
I tic on the tiniest most useless details, I'm sorry for bringing it up! I just can't help it I wanna know why ^^

PsiACE:
I think this is a great refinement of attitude and thank you for your PRs.

moodule:
Thanks for checking them!

This is assembly line work for me, I forgot I was talking to a human and wrote these like a random commit, sorry !

I'm rewrapping my customisation of this repo and seeing if anything is useful, I stacked a few commits over the year.

the irony is I dared to add a "code of conduct" :D

PR #544, opened 2019-11-24
Removed all references to `setup.py test` fixes #500

### LINKED ISSUE #500: `setup.py test` is bad practice
`setup.py test` is deprecated and will be removed one day: https://github.com/pypa/setuptools/issues/1684
`pytest-runner` is no longer recommended: https://github.com/pytest-dev/pytest-runner/issues/50

The project should be updated to avoid these "smells".

ChadBailey:
Note that this was done as a separate bugfix and does not include the linting and test fixes in my prior PR. I can resubmit with both integrated if needed

### DIFF
--- a/tests/test_bake_project.py
+++ b/tests/test_bake_project.py
@@ -90,7 +90,7 @@ def test_bake_with_defaults(cookies):
 def test_bake_and_run_tests(cookies):
     with bake_in_temp_dir(cookies) as result:
         assert result.project.isdir()
-        run_inside_dir('python setup.py test', str(result.project)) == 0
+        run_inside_dir('python -m unittest discover', str(result.project)) == 0
         print("test_bake_and_run_tests path", str(result.project))
 
 
@@ -98,14 +98,14 @@ def test_bake_withspecialchars_and_run_tests(cookies):
     """Ensure that a `full_name` with double quotes does not break setup.py"""
     with bake_in_temp_dir(cookies, extra_context={'full_name': 'name "quote" name'}) as result:
         assert result.project.isdir()
-        run_inside_dir('python setup.py test', str(result.project)) == 0
+        run_inside_dir('python setup.py --help', str(result.project)) == 0
 
 
 def test_bake_with_apostrophe_and_run_tests(cookies):
     """Ensure that a `full_name` with apostrophes does not break setup.py"""
     with bake_in_temp_dir(cookies, extra_context={'full_name': "O'connor"}) as result:
         assert result.project.isdir()
-        run_inside_dir('python setup.py test', str(result.project)) == 0
+        run_inside_dir('python setup.py --help', str(result.project)) == 0
 
 
 # def test_bake_and_run_travis_pypi_setup(cookies):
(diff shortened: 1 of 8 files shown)


tests/test_bake_project.py
```diff
@@ -90,22 +90,22 @@ def test_bake_with_defaults(cookies):
...
+        run_inside_dir('python -m unittest discover', str(result.project)) == 0
         print("test_bake_and_run_tests path", str(result.project))
 
 
 def test_bake_withspecialchars_and_run_tests(cookies):
     """Ensure that a `full_name` with double quotes does not break setup.py"""
     with bake_in_temp_dir(cookies, extra_context={'full_name': 'name "quote" name'}) as result:
         assert result.project.isdir()
-        run_inside_dir('python setup.py test', str(result.project)) == 0
+        run_inside_dir('python setup.py --help', str(result.project)) == 0
```
jamesmyatt: This doesn't do what the test name says

tests/test_bake_project.py
```diff
@@ -90,22 +90,22 @@ def test_bake_with_defaults(cookies):
...
 
 
 def test_bake_with_apostrophe_and_run_tests(cookies):
     """Ensure that a `full_name` with apostrophes does not break setup.py"""
     with bake_in_temp_dir(cookies, extra_context={'full_name': "O'connor"}) as result:
         assert result.project.isdir()
-        run_inside_dir('python setup.py test', str(result.project)) == 0
+        run_inside_dir('python setup.py --help', str(result.project)) == 0
```
jamesmyatt: This doesn't do what the test name says

{{cookiecutter.project_slug}}/setup.cfg
```diff
@@ -19,10 +19,3 @@ exclude = docs
 
 [aliases]
```
jamesmyatt: I'd remove this section if it's empty

{{cookiecutter.project_slug}}/tox.ini
```diff
@@ -17,15 +17,15 @@ commands = flake8 {{ cookiecutter.project_slug }}
...
-{% if cookiecutter.use_pytest == 'y' -%}
 deps =
     -r{toxinidir}/requirements_dev.txt
 ; If you want to make tox run the tests with the same versions, create a
 ; requirements.txt with the pinned versions and uncomment the following line:
 ;     -r{toxinidir}/requirements.txt
+{% if cookiecutter.use_pytest == 'y' -%}
 commands =
     pip install -U pip
     pytest --basetemp={envtmpdir}
```
jamesmyatt: Above, the command was `pytest ./tests`

tests/test_bake_project.py
ChadBailey: You're right... When I saw the docstring I rationalized that the only reason it was running the tests was to validate that setup.py was formatted correctly. Do you agree/does this accomplish the same objective with less overhead? If so I'll rename it, if not I'll change it to run tests.

tests/test_bake_project.py
ChadBailey: see previous comments

{{cookiecutter.project_slug}}/tox.ini
ChadBailey: I moved this down because requirements_dev.txt needs to be installed irrespective of which test suite is in use.

{{cookiecutter.project_slug}}/setup.cfg
ChadBailey: I left this in to demonstrate how it's done, but I can certainly remove it

PR #562, opened 2020-02-05
Ignoring venv directory in command find

williamcanin:
When searching with find, it is looking in the venv directory (if any) and trying to remove folders with line 36.
When line 36 is executed, there is a failure to find folders containing ".egg", and when trying to remove the folders, the failure is thrown because it does not have the "r" flag in "rm -fr".

### DIFF
--- a/{{cookiecutter.project_slug}}/Makefile
+++ b/{{cookiecutter.project_slug}}/Makefile
@@ -32,14 +32,14 @@ clean-build: ## remove build artifacts
 	rm -fr build/
 	rm -fr dist/
 	rm -fr .eggs/
-	find . -name '*.egg-info' -exec rm -fr {} +
-	find . -name '*.egg' -exec rm -f {} +
+	find . '!' -path './venv/*' -name '*.egg-info' -exec rm -fr {} +
+	find . '!' -path './venv/*' -name '*.egg' -exec rm -f {} +
 
 clean-pyc: ## remove Python file artifacts
-	find . -name '*.pyc' -exec rm -f {} +
-	find . -name '*.pyo' -exec rm -f {} +
-	find . -name '*~' -exec rm -f {} +
-	find . -name '__pycache__' -exec rm -fr {} +
+	find . '!' -path './venv/*' -name '*.pyc' -exec rm -f {} +
+	find . '!' -path './venv/*' -name '*.pyo' -exec rm -f {} +
+	find . '!' -path './venv/*' -name '*~' -exec rm -f {} +
+	find . '!' -path './venv/*' -name '__pycache__' -exec rm -fr {} +
 
 clean-test: ## remove test and coverage artifacts
 	rm -fr .tox/


PsiACE:
I see your question #564, I agree with you, but the problem exists in multiple directories.

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -32,14 +32,14 @@ clean-build: ## remove build artifacts
 	rm -fr build/
 	rm -fr dist/
 	rm -fr .eggs/
-	find . -name '*.egg-info' -exec rm -fr {} +
-	find . -name '*.egg' -exec rm -f {} +
+	find . '!' -path './venv/*' -name '*.egg-info' -exec rm -fr {} +
+	find . '!' -path './venv/*' -name '*.egg' -exec rm -f {} +
```
PsiACE: I think we may need to ignore the following directories.

```
# virtualenv
env/
.env/
.venv
venv/
ENV/
```

{{cookiecutter.project_slug}}/Makefile
```diff
@@ -32,14 +32,14 @@ clean-build: ## remove build artifacts
...
 
 clean-pyc: ## remove Python file artifacts
-	find . -name '*.pyc' -exec rm -f {} +
-	find . -name '*.pyo' -exec rm -f {} +
-	find . -name '*~' -exec rm -f {} +
-	find . -name '__pycache__' -exec rm -fr {} +
+	find . '!' -path './venv/*' -name '*.pyc' -exec rm -f {} +
+	find . '!' -path './venv/*' -name '*.pyo' -exec rm -f {} +
+	find . '!' -path './venv/*' -name '*~' -exec rm -f {} +
+	find . '!' -path './venv/*' -name '__pycache__' -exec rm -fr {} +
```
PsiACE: Do I need to ignore the `virtualenv` directory here?

williamcanin: Yes! One way I thought is to create a list of directories and file extensions to be excluded, ignoring the virtualenv directories.
So, to plan in the **Makefile** a "**define**" to incorporate the **Python** code.
I created a generic code that cleans up all the unwanted files. In my opinion, this is much more interesting, making Makefile options more compact.

Code in Makefile:
```
define CLEAN_ALL_TRASH
import os
import shutil
ROOT = os.getcwd()
ignore_dirs = ['venv', '.venv', 'env', '.env', 'ENV']
exclude_files = ['.egg', '.pyc', '*~', '.pyo']
exclude_dirs = ['.egg-info', '__pycache__']
for root, dirs, files in os.walk(ROOT, topdown=True):
    dirs[:] = [d for d in dirs if d not in ignore_dirs]
    for dir_ in dirs:
        for i in exclude_dirs:
            if dir_.endswith(i):
                print('Removed the directory: ', os.path.join(root, dir_))
                shutil.rmtree(os.path.join(root, dir_))
    files = [file for file in files if os.path.splitext(file)[1] in exclude_files]
    for file in files:
        print('Removed the file: ', os.path.join(root, file))
        os.remove(os.path.join(root, file))
endef
export CLEAN_ALL_TRASH

clean: ## remove all artifacts
    @python -c "$$CLEAN_ALL_TRASH"
```

PsiACE:
Good job. But maybe we can:

```
	find . \( -path ./env -o -path ./venv -o -path ./.env -o -path ./.venv \) -prune -o -name '*.egg-info' -exec rm -fr {} +
	find . \( -path ./env -o -path ./venv -o -path ./.env -o -path ./.venv \) -prune -o -name '*.egg' -exec rm -f {} +
```

williamcanin:
> Good job. But maybe we can:
> 
> ```
> 	find . \( -path ./env -o -path ./venv -o -path ./.env -o -path ./.venv \) -prune -o -name '*.egg-info' -exec rm -fr {} +
> 	find . \( -path ./env -o -path ./venv -o -path ./.env -o -path ./.venv \) -prune -o -name '*.egg' -exec rm -f {} +
> ```

Yes, that was how I was thinking, just using command shell.

PR #566, opened 2020-02-08
Add minimal configuration file for Read the Docs

### LINKED ISSUE #411: Consider adding .readthedocs.yml
Hi! Read the Docs prefers the usage of a `yaml` file to configure your project under the platform instead of just the web admin interface. So, I think it would be good to add the YAML template file to this project.

Docs: http://docs.readthedocs.io/en/latest/yaml-config.html

> Although it says that's in beta, it's very robust and solid. I think that message is there because it's not possible _to configure all of the web settings_ by that file yet.

humitos:
This is the minimal configuration file for Read the Docs Config File V2 --which is the recommended way to setup a project on Read the Docs. See https://docs.readthedocs.io/en/stable/config-file/v2.html

There are other options that can be added by default if you want, but I think this is a good first version.

In summary, this file will tell Read the Docs to use:

- Sphinx with the `docs/conf.py` file
- Python 3 by default
- Build HTML format only (no ePUB, no PDF, no HTMLZip)
- Install default Read the Docs requirements to build the docs (Sphinx, etc)
- Install the Python package itself (required by `sphinx.ext.autodoc`)

Other things to consider:

- Use [Read the Docs theme](https://sphinx-rtd-theme.readthedocs.io/en/stable/) instead of Alabaster
- Use [`sphinx-autoapi`](https://sphinx-autoapi.readthedocs.io/en/latest/) instead of `autodoc` to avoid installing dependencies to document the API
- Ask at startup for documentation output formats
- Ask at startup if the package needs `conda` to create the environment instead of regular virtualenv its docs

Let me know what you think!

Closes #411

### DIFF
--- a/{{cookiecutter.project_slug}}/.readthedocs.yml
+++ b/{{cookiecutter.project_slug}}/.readthedocs.yml
@@ -1,0 +1,15 @@
+# .readthedocs.yml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+
+# Set requirements required to build your docs
+python:
+  - method: pip
+    path: .


PsiACE:
I like your work very much, but there are some issues to consider:

- This project itself should also use this configuration file.
- I still think it is necessary to use regular virtualenv.

Other ideas are great, look forward to seeing your further work.

PR #571, opened 2020-02-19
Keep Dask fragments out of commits

accurrently:
The folder `dask-worker-space/` is a cache folder (like `__pycache__/`) for projects that implement [Dask](https://dask.org). If a data science or distributed computing project package uses Jupyter notebooks (e.g. in an `examples/` folder), there is a chance that they might use Dask. Similarly, tests for modules that implement Dask will create a `dask-worker-space/` folder as well if the code is run. This addition keeps that cache out of commits.

See [Dask on Github](https://github.com/dask/dask) repo for project information.


pydanny (approved):
LGTM!

PR #576, opened 2020-03-29
Introduces Black linter (disabled by default), closes #575

### LINKED ISSUE #575: Plans to use black? Are pull requests welcome?
Is there plan to use black https://github.com/psf/black ?
If so are pull requests integrating it welcome?

AndreMiras:
This is a first iteration with things kept simple. In the future I think we should do things like actually running both linters (flake and black) on the generated project via `tests/test_bake_project.py`.
Among other things I can also later provide a `make format` target that automatically fixes linter issues.
If things go as planned we can also add the `isort` linting option too

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -8,6 +8,7 @@
   "pypi_username": "{{ cookiecutter.github_username }}",
   "version": "0.1.0",
   "use_pytest": "n",
+  "use_black": "n",
   "use_pypi_deployment_with_travis": "y",
   "add_pyup_badge": "n",
   "command_line_interface": ["Click", "Argparse", "No command-line interface"],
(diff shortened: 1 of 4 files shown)


hXtreme:
Just pinging here because there seems to be a problem with TravisCI. [Here](https://travis-ci.org/github/audreyr/cookiecutter-pypackage/builds/668491251) it says the build has passed but github says that the build is pending.

@AndreMiras you might wanna push an empty commit or force-push to trigger the CI again and hopefully fix this.

_(I really like black and would love to see this merged soon)_

hrishikeshrt:
Any updates on this?

AndreMiras:
I've updated the pull request fixing merge conflicts and bumped black to the last version

PsiACE:
thanks, i will merge it

PR #578, opened 2020-03-29
Simplifies PyPI deployment over Travis configuration

AndreMiras:
Leverages the not so well documented token and `PYPI_PASSWORD` feature.
It's actually already possible in Travis v1 to use this environment
variable refs:
https://github.com/travis-ci/dpl/blob/v1.10.15/lib/dpl/provider/pypi.rb#L12
Note that the `__token__` user is a PyPI feature, refs:
https://pypi.org/help/#apitoken
This is more secure than user and password as tokens are generated with
uploaded only permissions and can be scoped to a given project.

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -5,7 +5,6 @@
   "project_name": "Python Boilerplate",
   "project_slug": "{{ cookiecutter.project_name.lower().replace(' ', '_').replace('-', '_') }}",
   "project_short_description": "Python Boilerplate contains all the boilerplate you need to create a Python package.",
-  "pypi_username": "{{ cookiecutter.github_username }}",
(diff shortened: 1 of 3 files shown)


browniebroke:
I've tried to use this after generating a project with this cookiecuter template, but it didn't work at first, it was showing up an error with `missing PyPI password` despite me adding the environment variable.

The trick was that the `PYPI_PASSWORD` environment variable was set to be available on `master` branch only, but Travis counts tags build to be on their own branch. 

I've than set the `PYPI_PASSWORD` to be avaiable on all branches, and now it works fine.

pydanny:
@AndreMiras Thanks so much for this contribution but we're moving the project off Travis.

PR #579, opened 2020-03-29
Adds missing asserts to `run_inside_dir() == 0`, closes #449

AndreMiras:
This is a follow up for #449 
There's no reason to compare with `== 0` if there's no assert.
I case we actually don't want to compare with zero, the next PR #580 covers that too


audreyfeldroy:
Thanks for the contribution! This project has been significantly modernized since this PR was opened (now uses pyproject.toml, uv, GitHub Actions, Typer). The files and features this PR touches no longer exist in the current template. Closing as obsolete.

Issue #670, opened 2022-08-28
Test fail when argument order are reversed in config


mazzma12:
* Date you used Cookiecutter PyPackage: 28 August 2022
* Cookiecutter version used, if any: 1.7.3
* Python version, if any: 3.9
* Operating System: Ubuntu 22.04

### Description

I am trying to add new tests and I noticed that swapping the argument order in the init config will let the current test fail. Do you know why ? I don't understand since it does not look than anything is built based on argument order (to me)
### What I Did

```
sed -i 's/"Click", "Argparse"/"Argparse", "Click"/g' ./cookiecutter.json
```

Then running the test produce 2 tests to fail:

```python
self = <click.testing.CliRunner object at 0x7f77df476e90>, cli = <function main at 0x7f77df51ca60>

    def get_default_prog_name(self, cli: "BaseCommand") -> str:
        """Given a command object it will return the default program name
        for it.  The default is the `name` attribute or ``"root"`` if not
        set.
        """
>       return cli.name or "root"
E       AttributeError: 'function' object has no attribute 'name'

../../../.pyenv/versions/cs310/lib/python3.10/site-packages/click/testing.py:195: AttributeError
=============================================================== short test summary info ===============================================================
FAILED tests/test_bake_project.py::test_bake_with_console_script_cli - AttributeError: 'function' object has no attribute 'name'
FAILED tests/test_bake_project.py::test_bake_with_argparse_console_script_cli - AttributeError: 'function' object has no attribute 'name
```

mazzma12:
refer to #67

audreyfeldroy:
@mazzma12 thanks a ton for catching this, and for the detailed issue report and PR :pray:

audreyfeldroy:
@CodiumAI-Agent /describe

audreyfeldroy:
@CodiumAI-Agent /ask What do you mean by *handle the case sensitivity issue within the code itself*?

mazzma12:
> @mazzma12 thanks a ton for catching this, and for the detailed issue report and PR 🙏

You welcome ! (Silently watching AI at work :smile: )

mazzma12:
> @mazzma12 thanks a ton for catching this, and for the detailed issue report and PR 🙏

You welcome ! (Silently watching AI at work :smile: )

PR #671, opened 2022-09-05
Fix Issue #670 : bake context argument should match the case 

mazzma12:
To my understanding, this would solve the issue describe in #670. It fallback to the first item in the list if the case is not respected.  It becomes a problem then if you try swapping argument order.

Also, this change likely causes the argparse test to fail, but this is maybe expected as you are testing  non Click input with the Click RunnerCli

### DIFF
--- a/tests/test_bake_project.py
+++ b/tests/test_bake_project.py
@@ -269,7 +269,7 @@ def test_bake_with_no_console_script(cookies):
 
 
 def test_bake_with_console_script_files(cookies):
-    context = {'command_line_interface': 'click'}
+    context = {'command_line_interface': 'Click'}
     result = cookies.bake(extra_context=context)
     project_path, project_slug, project_dir = project_info(result)
     found_project_files = os.listdir(project_dir)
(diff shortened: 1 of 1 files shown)


audreyfeldroy:
@CodiumAI-Agent /review

audreyfeldroy:
@CodiumAI-Agent /improve

audreyfeldroy:
@CodiumAI-Agent /describe

audreyfeldroy:
@CodiumAI-Agent /ask What do you mean by *handle the case sensitivity issue within the code itself*?

audreyfeldroy:
@CodiumAI-Agent /ask Where is "the code that processes the 'command_line_interface' value"?

audreyfeldroy:
@CodiumAI-Agent /ask Find the code that processes the 'command_line_interface' value in https://github.com/hackebrot/pytest-cookies/blob/main/src/pytest_cookies/plugin.py - possibly in the 'cookies.bake' method.

Issue #735, opened 2024-04-06
Better filtering on package data


pydanny:
### Description

Right now we just include everything, but I think we can filter out some unnecessary files.

PR #736, opened 2024-04-06
Add missing package data

pydanny:
Not having this means namespaces aren't working because there are files not being added into the namespace. Which means tests inside the rendered package fails. We will address this in #735.

https://setuptools.pypa.io/en/latest/userguide/datafiles.html


FrancisBilla (approved):
Look good to me

PR #854, opened 2025-01-15
Add __main__.py

triveria:
Hi,

on creating a python program I noticed I couldn't execute it since it didn't have a `__main__.py`.
Furthermore, I also couldn't start the debugger on that package for the very same reason.
 
This PR adds the missing `__main__.py` to the template to make it possible to both execute and debug the generated package.

I've seen [PR #481](https://github.com/audreyfeldroy/cookiecutter-pypackage/pull/481), where @znd4 already worked on a very similar PR which unfortunatelly was never merged.
Based on the coversation there, I've tried to pick up where he left.

Let me know if I'm missing anything here.

Sidenote: While adding this, I've noticed a couple things:
  - The unit tests seem to be broken.
  - When the user selects `"command_line_interface" = "Argparse`, the project still gets set up for `Typer`

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -10,6 +10,7 @@ if __name__ == '__main__':
 
     if 'no' in '{{ cookiecutter.command_line_interface|lower }}':
         pathlib.Path('src', '{{ cookiecutter.project_slug }}', 'cli.py').unlink()
-        
+        pathlib.Path('src', '{{ cookiecutter.project_slug }}', '__main__.py').unlink()
+
     if 'Not open source' == '{{ cookiecutter.open_source_license }}':
         pathlib.Path('LICENSE').unlink()
--- a/{{cookiecutter.project_slug}}/pyproject.toml
+++ b/{{cookiecutter.project_slug}}/pyproject.toml
@@ -18,7 +18,7 @@ classifiers = [
 ]
 license = {text = "{{cookiecutter.open_source_license}}"}
 dependencies = [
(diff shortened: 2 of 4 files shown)


CalOmnie:
Hey, It would probably be better to simply rename `cli.py` to `__main__.py`. especially given that `cli.py` already has a `if __name__ == "__main__"` statement. Keeping it as is would mean that there would be 2 entrypoint in the project which is not ideal.

triveria:
I'd be happy to do that as it would also be my preferred solution, but there are two caveats:

- In [PR #481](https://github.com/audreyfeldroy/cookiecutter-pypackage/pull/481), the contributor initially wanted to rename `cli.py` to `__main__.py`, which was rejected with the notion that the package should have both a `__main__.py` and a `cli.py`.
- Some high-profile packages use both `cli.py` and `__main__.py`, which indicates that this is best practice. See, for example, [Flask](https://github.com/pallets/flask/tree/main/src/flask) and [Cookiecutter](https://github.com/cookiecutter/cookiecutter/tree/main/cookiecutter).

I'd like to have a verdict from a collaborator on this issue: @audreyfeldroy, @pydanny, @ArjunDandagi.

triveria:
Pinging @audreyfeldroy @pydanny 
Please review this PR. You have a great cookiecutter template going that I often use. But I see the last commit is almost a year old.
Please keep this template alive and merge some of the outstanding PRs.
It would be a real bummer if this project became outdated.

audreyfeldroy:
Thank you @triveria!

PR #857, opened 2025-02-04
Move away from TravisCI

### LINKED ISSUE #695: Travis not usable anymore
Travis CI is no longer available free. As this package is widely seen as the standard cookiecutter package, moving to another ci cd supplier shall be considered. Maybe githubs own GitHub Actions?

JostMigenda:
This closes #695 by removing all references to TravisCI and adding a GitHub Actions workflow for testing and linting instead.

A few notes:
* Instead of `tox`, this workflow uses GitHub Actions’ native capabilities for testing across multiple Python versions. This seems more straightforward and is more easily extensible e.g. to test on different operating systems. I’ve kept tox everywhere else, for local testing. (Personally, I’d be tempted to remove it completely—but that’s better left to a separate PR.)
* It is also possible to set up a GitHub Actions workflow to automatically publish to PyPI. GitHub docs have [a complete workflow template](https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#publishing-to-pypi) for that; for now, I’ve included a link to that in `CONTRIBUTING.rst`s “Deployment” section, but I could also add that workflow to the template directly, if you think that makes sense?


audreyfeldroy (approved):
Merging, thanks @JostMigenda!

PR #859, opened 2025-04-15
Set dynamic version

rafa-guedes:
This change allows versions to be set from the `__version__` attribute instead of manually defined in the pyproject config.


audreyfeldroy:
Thanks @rafa-guedes!

alanwilter:
Sorry, I don't dynamic version in the latest release 0.4.0. Why?

PR #867, opened 2025-08-25
Feature/build automation

nivika11:


audreyfeldroy:
Thanks for thinking about this! These changes mix a few different concerns and the Docker setup would belong in projects generated from this template rather than the template itself. Closing for now.

PR #870, opened 2026-02-03
Remove silent comparisons in test assertions

AndreMiras:
The run_inside_dir() function uses subprocess.check_call() which already raises CalledProcessError on non-zero exit codes. The == 0 comparisons were performed but never asserted, making them no-ops.

Closes #449


audreyfeldroy:
Merged via #875. Thanks again!

PR #906, opened 2026-02-18
Add groups and aliases to just recipes

npikall:
This PR 
- groups the available recipes when listed for a cleaner overview
- adds aliases to commonly used tasks
- and uses `uv` to show the projects version

As this template has been quite useful and inspiring to me, I wanted to show some gratitude and make it a bit more convenient to use.

### DIFF
--- a/{{cookiecutter.pypi_package_name}}/justfile
+++ b/{{cookiecutter.pypi_package_name}}/justfile
@@ -87,7 +87,7 @@ VERSION := `uv version`
 # Print the current version of the project
 [group("release")]
 version:
-    @echo "Current version: {{VERSION}}"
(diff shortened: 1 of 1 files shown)


audreyfeldroy:
Thank you for this! Merged. I especially like the `uv version` switch, that's much cleaner than the grep/sed parsing.

I made a small follow-up (70bf76c) removing the group annotations. I tend to read justfiles directly rather than through `just --list`, so I prefer keeping the file itself as clean as possible.

npikall:
@audreyfeldroy  Great! Yes that's very understandable, I just found them quite helpful for me when listing a lot of recipes. Anyway, I am still glad the aliases and the `uv version` are keepers.

PR #907, opened 2026-03-06
docs: Add CLI key=value override documentation to README

### LINKED ISSUE #879: Add README documentation for key=value CLI feature
## Summary

The CLI now supports passing `key=value` arguments to override template variables, but this feature isn't documented in the README.

## What's needed

Add a section to the README (probably after Quickstart) showing:

1. Basic example:
```bash
uvx cookiecutter-pypackage full_name="Your Name" github_username=yourhandle
```

2. Using with `--no-input` for automation:
```bash
uvx cookiecutter-pypackage --no-input full_name="Your Name" email="you@example.com"
```

3. Shell quoting note:
- `full_name="First Last"` (correct)
- `full_name=First Last` (incorrect, shell splits this)

4. List of available variables (or link to where they're documented)

## Priority

Medium. The feature works without docs, but docs would improve discoverability for automation use cases.

klouds27:
Closes #879

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,18 @@
 # Cookiecutter PyPackage
+
 [![PyPI version](https://img.shields.io/pypi/v/cookiecutter-pypackage.svg)](https://pypi.python.org/pypi/cookiecutter-pypackage)
 [![PyPI downloads](https://img.shields.io/pypi/dm/cookiecutter-pypackage.svg)](https://pypi.python.org/pypi/cookiecutter-pypackage)
+
 [Cookiecutter](https://github.com/cookiecutter/cookiecutter) template for a Python package with production-ready CI and automated PyPI publishing. Push a tag, your package ships, no token management needed.
+
 ```bash
 uvx cookiecutter-pypackage
 ```
+
 ## What you get
+
 ### Tooling
+
 | | Tool | |
 |---|---|---|
 | Package manager | [uv](https://docs.astral.sh/uv/) | Fast, handles venvs automatically |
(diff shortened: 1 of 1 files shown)


audreyfeldroy (changes requested):
Thanks for tackling this! The CLI override feature really does need docs. A few things to address before this can merge:

**Formatting regression:** The diff strips most blank lines between sections throughout the entire README. This is probably an editor config thing rather than intentional. Easiest path: start fresh from `main` and add only the new section. Also, the trailing newline at the end of the file got removed.

**Shell quoting note:** Issue #879 specifically called for this, and it's the thing most likely to actually save someone time:

```
full_name="First Last"   # correct
full_name=First Last      # wrong, shell splits this
```

**Example variables:** `project_slug` is derived from `pypi_package_name` in `cookiecutter.json`, so overriding it directly while leaving `pypi_package_name` at its default creates a mismatch. Better to show `pypi_package_name` and let the slug derive naturally. Also worth noting that any variable you don't pass gets its default from `cookiecutter.json`, so the `--no-input` example doesn't need to list everything.

audreyfeldroy:
The original review feedback was fully addressed at the time. Since then, the template schema changed, leaving the examples stale.

I rebased and updated the work in #932 while preserving @klouds27 as the author of the original documentation commit and crediting @iamamystery for the current-variable examples incorporated from #929/#930.

GitHub rejected authenticated maintainer pushes to this fork branch over both HTTPS and SSH despite maintainer edits being enabled, so the repository maintenance policy required an origin-hosted replacement. Closing this PR as superseded by #932.

PR #929, opened 2026-06-19
docs: add key=value CLI feature documentation to README

iamamystery:
Closes #879

## What
Adds documentation for the key=value CLI feature to README.md
as requested in #879.

## Checklist
- [x] Addresses exactly one issue or feature
- [x] Diff contains only changes for this task

@audreyfeldroy please review when you get a chance!

PR #930, opened 2026-06-19
feat: add --list-variables flag to show available template variables

iamamystery:
## AI Provenance

- **Tool:** 
- **Model:** 
- **Prompt/thread:** 

## Checklist

- [ ] Addresses exactly one issue or feature
- [ ] New or changed behavior has test coverage
- [ ] Diff contains only changes for this task

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,25 @@ Install [uv](https://docs.astral.sh/uv/getting-started/installation/), then:
 ```bash
 uvx cookiecutter-pypackage
 ```
+## Overriding Template Variables
+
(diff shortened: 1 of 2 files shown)


audreyfeldroy:
The README commit in this PR is identical to #929 and has been incorporated into #932 with Muhammad Jawad's contribution credited.

The Ruff `uvx` change is unrelated to #879 and is intentionally not carried over because it would bypass a committed `uv.lock` in generated projects. If CI installation performance is revisited, it should be a separate measured change that keeps the outer and generated workflows in sync.

Closing this mixed-scope duplicate in favor of #932.

